PR #4, opened 2018-07-12
D2l icon version

jwong-d2l:


components/d2l-sequences-iterator.html
```diff
@@ -1,95 +1,97 @@
...
+            }
+            static get properties() {
+                return {
+                    disabled: {
+                        type: Boolean,
+                        computed: '_isDisabled(entity)'
+                    },
+                    icon: {
+                        type: String,
+						reflectToAttribute: true
```
jbydeley-d2l: Doesn't this just populate the attribute if `icon` changes internally? Is that something that can happen?

jwong-d2l: I'm not sure, but without it, it wasn't working.  I'll test it again.

PR #7, opened 2018-07-19
Us98458/use d2l button for download page

tinglisD2L:


components/d2l-sequences-content-file-download.html
```diff
@@ -68,17 +65,19 @@
...
 			</div>
 			<div>
 				[[localize('fileSize','size',_fileSize)]]
 			</div>
 			<template is="dom-if" if="[[_fileLocation]]">
 				<form method="get" action$="[[_fileLocation]]">
-					<button type="submit" class="download-button heading3">
-						[[localize('download')]]
+					<button type="submit" class="download-button">
+						<d2l-button primary>
```
jwong-d2l: does the form attribute not do a submit?

tinglisD2L: what do you mean? what's the setup you're thinking of? i can test it

jwong-d2l: this guy: https://github.com/BrightspaceUI/button/blob/master/d2l-button.html#L87

does giving it the ID of your form make it submit (so you don't have to wrap d2l-button in a button?)

tinglisD2L: I tried giving the form id="myForm" and d2l-button form="myForm". That didn't work.
I tried giving d2l-button formmethod="get" formaction$="[[_fileLocation]]. Didn't work.
The combination of the two didn't work either.
:\

tinglisD2L: oh i didn't try adding type. let me try

tinglisD2L: nothing works. I don't know why

PR #9, opened 2018-07-19
US98671 TA126980 Introduce aria to iterator

lawong-d2l:
[RALLY](https://rally1.rallydev.com/#/15545269080ud/detail/userstory/237779530676)

- At first I was using aria-labels and doing crazy things with tabindex. I thought I had to somehow get around the nesting that these custom elements are build around the native `<button>` tags. Turns out I was just supposed to keep it [simple](https://github.com/BrightspaceUI/icons/issues/167)

- Updated tests and added some lang terms

- Works well with NVDA on both Firefox and Chrome


components/d2l-sequences-iterator.html
```diff
@@ -49,9 +53,38 @@
...
+				this.loadResources(this.resolveUrl('../locales.json'));
+                 if (this.next) {
+                    this.iterateTo = 'iterateToNext';
+                }
+                else if(this.previous) {
+                    this.iterateTo =  'iterateToPrevious';
+                }
+                else if(this.up) {
+                    this.iterateTo =  'iterateToParent';
+                }
```
RyanSweetD2L: I suspect that we want to move this logic into an iterateTo observer.  Leaving it defined here means that it is only evaluated on connection.  If someone removed a `next` attribute and added a `prev` attribute then it would act like a previous button but aria would still indicate it is a next button.

components/d2l-sequences-iterator.html
```diff
@@ -49,9 +53,34 @@
...
+				this.loadResources(this.resolveUrl('../locales.json'));    
+			}
+            getIterateTo() {
+                if (this.next) {
+                    return 'iterateToNext';
+                }
+                else if(this.previous) {
+                    return 'iterateToPrevious';
+                }
+                else if(this.up) {
```
jwong-d2l: for the iterators on our page, where is this one used?

lawong-d2l: Nowhere., it used to be the "back" button. We just left the capability for anyone that wants to use something like this.

jwong-d2l: that's what I thought.  thanks!

PR #12, opened 2018-07-20
Add completion tracking support

RyanSweetD2L:
- Call siren action 'view-activity-duration' when starting content items
- Call siren action 'finish-view-activity' on finish of viewed items

### DIFF
--- a/mixins/d2l-completion-tracking-mixin.html
+++ b/mixins/d2l-completion-tracking-mixin.html
@@ -1,0 +1,94 @@
+<link rel="import" href="../../polymer-siren-mixins/siren-action-mixin.html">
+<script>
+	(function() {
+		function CompletionTrackingMixin(superClass) {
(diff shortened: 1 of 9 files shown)


mixins/d2l-completion-tracking-mixin.html
```diff
@@ -0,0 +1,94 @@
...
+							computed: '_isImpersonating(token)'
+						}
+					};
+				}
+
+				static get observers() {
+					return ['_entityUpdated(entity)'];
+				}
+
+				disconnectedCallback() {
```
jwong-d2l: does this get called when the entity is updated?

jbydeley-d2l: It gets called when the the web component gets removed from the page. In this case, it's when we navigate away from one type of content to another.

PR #22, opened 2018-07-27
Reject a promise instead of returning null

RyanSweetD2L:


mixins/d2l-sequences-completion-tracking-mixin.html
```diff
@@ -30,24 +30,33 @@
 					}
 
 					this._performCompletion(this.entity, 'view-activity-duration')
-						.then(completion => this._completionEntity = completion);
+						.then(completion => this._completionEntity = completion)
+						.catch(() => {});
```
RyanSweetD2L: The browser complains if I don't include the catch handler even if I don't really want to do anything...

PR #25, opened 2018-08-14
Convert to using polymer-siren-behaviors from polymer-siren-mixins

RyanSweetD2L:
- Refactor main dependencies into completion-tracking-mixin, router-mixin
- Rename router-mixin to align with other sequence mixins
- Refactor to use entity methods (getLinkByRel) instead of mixin methods
  (_getLinkByRel)


test/d2l-sequences-automatic-completion-tracking-mixin.html
```diff
@@ -42,15 +42,11 @@
 				element = fixture('MixinFixture');
 			});
 
-			it('startCompletion must be called when element is updated', async() => {
-				const startCompletion = sinon.spy(element, 'startCompletion');
+			it('startCompletion and finishCompletion must be called when element is updated', async() => {
```
RyanSweetD2L: I combined these two tests because there were errors in the console.log when not using a stub implementation.  Since both stubs are always invoked it doesn't make sense to just test one without the other.

mixins/RouterMixin.html
```diff
@@ -1,60 +0,0 @@
-<script>
```
RyanSweetD2L: This file was renamed.  I'm not sure why the diff isn't treating it like that...

components/d2l-sequences-content-unknown.html
```diff
@@ -11,15 +8,9 @@
     </template>
     <script>
 		/*
-			@mixes SirenEntityMixin
             @extends D2L.PolymerBehaviors.Sequences.LocalizeBehavior
```
jwong-d2l: does this tag need to change?

PR #28, opened 2018-08-20
Optimize the content router

RyanSweetD2L:
- Chaining multiple routers caused multiple DOM nodes to be created
- Unifying the router allows us to lookup the activity subentity once
  and act on it immediately


test/d2l-sequences-content-router.html
```diff
@@ -19,16 +19,64 @@
 		<script>
 			/* global loadFileWithDelay */
 			describe('d2l-sequences-content-router', () => {
-				it('should create a d2l-sequences-content-file-router for file-activity types', async() => {
-					const element = await  loadFileWithDelay('data/activity-file-html.json', fixture('RouterFixture'));
-					expect(element.shadowRoot.querySelector('d2l-sequences-content-file-router')).to.exist;
-					expect(element.shadowRoot.querySelector('d2l-sequences-content-unknown')).not.to.exist;
+				describe('should handle a file activity for the mimetype:', () => {
+					it('application/pdf', async() => {
```
tinglisD2L: nit: The descriptions for `application/pdf`, `image/bmp` and `text/html` could be rewritten as 'should handle {file-type}', if you want.

PR #29, opened 2018-08-28
scroll to top fix

tinglisD2L:


tinglisD2L:
@jbydeley-d2l Why IE11? I thought our app doesn't work on IE11 at all and it's not supported. I can try Edge though.

PR #30, opened 2018-09-04
fix bug with scrollToTop function not defined

tinglisD2L:
https://rally1.rallydev.com/#/15545269080/detail/task/251063130724

### DIFF
--- a/components/d2l-sequences-content-link.html
+++ b/components/d2l-sequences-content-link.html
@@ -37,6 +37,9 @@
 					_linkLocation: {
 						type: String,
 						computed: '_getLinkLocation(entity)'
+					},
+					_loadListener: {
+						type: Object
 					}
 				}
 			}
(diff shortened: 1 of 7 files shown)


components/d2l-sequences-content-link.html
```diff
@@ -2,73 +2,83 @@
...
+			static get observers() {
+				return ['_render(entity)']
+			}
 
-                // initially hide the content, show the spinner
-                content.classList.add('hide');
-                spinner.classList.remove('hide');
+			ready() {
+				super.ready();
+				this.$.content.addEventListener('load', this._scrollToTop.bind(this));
```
jbydeley-d2l: I'm not 100% certain but this looks suspicious to me. With event listeners we often need to remove them if we are finished with them. I'd feel safer if we went with the pattern in [d2l-sequence-viewer](https://github.com/Brightspace/d2l-sequence-viewer/blob/master/d2l-sequence-viewer.html#L191) with `on-blur`

tinglisD2L:
ok tested. it's all good

PR #31, opened 2018-09-10
Tinglis/add end of lesson pages

tinglisD2L:
Related PRS:
https://github.com/Brightspace/d2l-sequence-viewer/pull/76
https://github.com/BrightspaceHypermediaComponents/d2l-sequence-navigator/pull/54


components/d2l-sequences-content-router.html
```diff
@@ -49,6 +53,8 @@
 							return getFileEntityType(subEntity);
 						case D2LSequencesContentRouter.linkActivity:
 							return getLinkEntityType(subEntity);
+						case D2LSequencesContentRouter.endOfLessonActivity:
+							return getEndOfLessonType(subEntity);
```
jbydeley-d2l: It seems like perhaps we should just `return D2LSequencesContentEoLMain.is;`

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+					</div>
+				</d2l-button>
+			</template>
+		</div>
+	</template>
+	<script>
+		/*
+			@extends D2L.PolymerBehaviors.Sequences.LocalizeBehavior
+		*/
+		class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.CompletionTrackingMixin() {
```
jbydeley-d2l: Do we track completion of our End of Lessons page?

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				margin: 30px 8px;
+				padding: 0px;
+				border: none;
+				background: transparent;
+			}
+		</style>
+		<div class="container">
+			<div class="image"></div>
+
+			<template is="dom-if" if="[[!hasMissing]]">
```
jbydeley-d2l: It feels like there should be two `dom-if`s and everything should be in one or the other. It would simplify readability if nothing else.

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				return completionCount.total - completionCount.completed;
+			}
+			_onclick() {
+				const location = this._getLinkLocation(this.entity);
+				if (!location) {
+					return;
+				}
+
+				this.startCompletion();
+				return window.open(location);
```
jbydeley-d2l: I had no idea about `window.open` and just read up on it after seeing this. Neat!

components/d2l-sequences-content-router.html
tinglisD2L: Oh yes. There were originally two types but I reduced them down to one.

components/d2l-sequences-content-eol-main.html
tinglisD2L: It uses the mixin to figure out whether to show the Congratulations page (if all completed) or the Good work so far page (if some are incomplete)

bower.json
```diff
@@ -19,6 +19,7 @@
     "d2l-icons": "#4.14.2",
```
dbatiste: Looks like the version could be updated to the latest...?  Do you know why this is being referenced differently?

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+<link rel="import" href="../../d2l-link/d2l-link.html">
+<link rel="import" href="../mixins/d2l-sequences-completion-tracking-mixin.html">
+<link rel="import" href="d2l-sequences-content-link.html">
+
+<dom-module id="d2l-sequences-content-eol-main">
+	<template>
+		<style>
+			:host {
+				display: block;
+				font-family: 'Lato', sans-serif;
```
dbatiste: Looks like extraneous typography styles here.

components/d2l-sequences-content-eol-main.html
tinglisD2L: completely agree! I don't know what prompted me to do it the other way to begin with!

components/d2l-sequences-content-eol-main.html
tinglisD2L: when I took that line away, the font style became a serif

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				color: var(--d2l-color-ferrite);
+
+				/* Body Text - Lato Regular 19 */
+				font-size: 19px;
+				line-height: 24px;
+				font-weight: 400; /* normal */
+				letter-spacing: 0.2px;
+			}
+
+			.container {
```
dbatiste: Note: generic class names in the component have a higher chance to cause styles bleed if something else on the page has defined this class with some other styles.  While shadow-DOM would isolate these nodes, shady-DOM will not.  So, our general practice for internal class names is... `component-*`.  In this case, `d2l-sequences-content-eol-main-container`.  The class names can sometimes get verbose and wordy, and we make effort to keep them concise.

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				text-align: center;
+				margin: auto;
+			}
+
+			.image {
+				width: 514px;
+				height: 289px;
+				background-color: #d8d8d8;
```
dbatiste: It looks like these colors are not in our color palette.  I am guessing this is from previous implementation?  Normally we would want to reference colors from [d2l-colors](https://github.com/BrightspaceUI/colors/blob/master/d2l-colors.html).

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				border: solid 1px #979797;
+				margin: 68px auto;
+			}
+
+			.heading2 {
+				/* Heading 2 - Lato Bold 30 */
+				font-size: 30px;
```
dbatiste: Note: there is a [--d2l-heading-2](https://github.com/BrightspaceUI/typography/blob/master/d2l-typography-shared-styles.html#L34) mixin that can be used.  Possible to switch?

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,195 @@
...
+				line-height: 36px;
+				font-weight: 400; /* normal */
+				letter-spacing: 0.2px;
+			}
+
+			.label-text {
+				/* Label Text - Lato Bold 14 */
```
dbatiste: Is it possible to use the [--d2l-label-text](https://github.com/BrightspaceUI/typography/blob/master/d2l-typography-shared-styles.html#L55) mixin?

components/d2l-sequences-content-eol-main.html
dbatiste: Is the app pulling in our typography from the LMS?  The host should be providing the typography.

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,186 @@
...
+				padding: 0px;
+				border: none;
+				background: transparent;
+			}
+		</style>
+		<div class="container">
+			<div class="image"></div>
+
+			<template is="dom-if" if="[[!hasMissing]]">
+				<div class="heading2 title">
```
dbatiste: Seems like the semantics are incorrect here - i.e. styling a `div` to be like a heading.  Headings have semantics that assistive tech can use for navigating the page.

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,186 @@
...
+
+			.title {
+				margin: 25px 0px;
+			}
+
+			.content {
+				margin: 5px 0px;
+			}
+
+			.button {
```
dbatiste: Curious what the purpose of restyling `d2l-button`?

bower.json
tinglisD2L: @jwong-d2l said it would break otherwise, and a new component is being created to address some issues. He would have the details if you're curious.

components/d2l-sequences-content-eol-main.html
```diff
@@ -0,0 +1,186 @@
...
+					},
+					hasMissing: {
+						Boolean,
+						computed: '_getHasMissing(missingCount)'
+					}
+				}
+			}
+			connectedCallback() {
+				super.connectedCallback();
+				this._language = window.document.getElementsByTagName('html')[0]
```
dbatiste: I am curious if our localization lib can handle this (I thought it would have).

bower.json
jwong-d2l: one of the components uses d2l-icon-button to have a navigation button that we can specify icon tier on.  The new one doesn't allow for this.

components/d2l-sequences-content-eol-main.html
tinglisD2L: I want to make the switch but --d2l-heading-2 looks smaller than what's in our design specs. I think it's because the mixin uses rem while I'm currently using px for font-size. I can't figure out how to scale everything up.

components/d2l-sequences-content-eol-main.html
tinglisD2L: Oh this doesn't matter. It's just a placeholder grey box until we get the actual image from design.

components/d2l-sequences-content-eol-main.html
tinglisD2L: Again, it's to fit with the design I'm given. It's mostly a sizing issue. If I could figure out the sizing issue regarding the font sizes, then this hack could probably be removed as well.

components/d2l-sequences-content-eol-main.html
tinglisD2L: I'm not sure. I copied it from another similar page.

dbatiste: It sounds like our localization lib does a bunch of things (fallback to org default locale, as well as timezone and other local overrides).  It seems like this should not be necessary - it would be good to try removing it, and if there's an issue, we can figure out what needs to change in the lib.

components/d2l-sequences-content-eol-main.html
dbatiste: The mixins in [d2l-typography](https://github.com/BrightspaceUI/typography) should match [the design patterns](http://design.d2l/style/typography/).  Hopefully the app design for the tool follows these patterns.  Apps shouldn't be overriding these unless it's a special scenario.  Otherwise we end up with a lot of inconsistency across the LMS, and future maintenance/updates are that much more difficult.

Is the goal of this work to get functional parity, or is there also effort to align with Daylight.

Note: the app (if configured correctly) should have a root font-size set, and the value is determined by the user's Account Settings.  The default is 20px, so when the mixin is used, the resulting font-size is 30px.  If the user has selected a different size through Account Settings, the resulting root font-size will be larger or smaller, and the resulting heading will be smaller or larger as well.

components/d2l-sequences-content-eol-main.html
tinglisD2L: @dbatiste I'm going to try to make the localization and typography changes now and once it's PR ready, I'll let you know

PR #32, opened 2018-09-10
Dispatch a toast event when navigating away from unsubmitted assignments

jbydeley-d2l:


jwong-d2l:
Correct me if I'm wrong, but this works for all activities?

jbydeley-d2l:
@jwong-d2l That's the plan. Right now it's looking at Dropbox but it could work for any of them.

PR #37, opened 2018-09-17
Tinglis/end of lesson attempt2

tinglisD2L:


components/d2l-sequences-content-link-scorm.html
```diff
@@ -16,30 +16,30 @@
...
 		<div class="content-link-scorm-container">
-			<div class="icon">
+			<div class="content-link-scorm-container-icon">
 				<d2l-scorm-icon></d2l-scorm-icon>
 			</div>
 			<p>
 				[[localize('viewThisActivity')]]
 			</p>
-			<button class="button" onClick$="window.open('[[_linkLocation]]')">
+			<button onClick$="window.open('[[_linkLocation]]')">
```
dbatiste: This is a bit odd - a button inside of a button.  Looks like this onclick should be moved onto `d2l-button`?

components/d2l-sequences-content-file-download.html
```diff
@@ -37,15 +37,15 @@ <h3>
...
+			<div class="content-file-download-container-file-icon">
 				<d2l-content-icons icon="file"></d2l-content-icons>
 			</div>
 			<p>
 				[[localize('fileSize','size',_fileSize)]]
 			</p>
 			<template is="dom-if" if="[[_fileLocation]]">
 				<form method="get" action$="[[_fileLocation]]">
-					<button type="submit" class="button">
+					<button type="submit">
```
lawong-d2l: is this a place we should be using d2l-button?

components/d2l-sequences-content-router.html
```diff
@@ -40,6 +41,9 @@
 			if (!entity || !entity.entities) {
 				return D2LSequencesContentUnknown.is;
 			}
+			if (entity.hasClass('end-of-sequence')) {
```
lawong-d2l: Don't think we need this bit anymore right?

components/d2l-sequences-content-router.html
```diff
@@ -1,4 +1,5 @@
 <link rel="import" href="../mixins/d2l-sequences-router-mixin.html">
+<link rel="import" href="d2l-sequences-content-eol-main.html">
```
lawong-d2l: Same for this, being not needed

components/d2l-sequences-content-file-download.html
tinglisD2L: I had trouble trying to get d2l-button to work with a form

components/d2l-sequences-content-router.html
tinglisD2L: sorry that's actually removed in the next PR that targets this branch. I wanted to keep the two PRs separate because they are separate stories but yes there were a few things I had to undo in the followup PR.

PR #38, opened 2018-09-18
Tinglis/add missed activities

tinglisD2L:


components/d2l-sequences-content-eol-activity-link.html
```diff
@@ -0,0 +1,92 @@
...
+					title: {
+						type: Object,
+						computed: '_getTitle(entity)'
+					},
+					href: {
+						type: String,
+						reflectToAttribute: true,
+						notify: true
+					},
+					token: {
```
lawong-d2l: Could be wrong I don't think it's necessary for us to be defining token, when it is done in the EntityBehavior

components/d2l-sequences-content-eol-main.html
```diff
@@ -82,9 +94,17 @@ <h2>
 			}
 			static get properties() {
 				return {
-					_location: {
+					href: {
+						type: String,
+						reflectToAttribute: true,
+						notify: true
+					},
+					token: {
```
lawong-d2l: token defined again?

components/d2l-sequences-content-eol-missed.html
```diff
@@ -0,0 +1,118 @@
...
+				return 'd2l-sequences-content-eol-missed';
+			}
+			static get properties() {
+				return {
+					href: {
+						type: String,
+						reflectToAttribute: true,
+						notify: true
+					},
+					token: {
```
lawong-d2l: token defined again?

components/d2l-sequences-content-eol-activity-link.html
tinglisD2L: @lawong-d2l When I didn't have these before, eslint was failing. Not sure why though

PR #41, opened 2018-09-24
add oneDrive page

tinglisD2L:


jwong-d2l:
What do you think about adding a generic external launch page and passed in icons and text?

components/d2l-sequences-content-link-onedrive.html
```diff
@@ -0,0 +1,78 @@
...
+			static get properties() {
+				return {
+					_linkLocation: {
+						type: String,
+						computed: '_getLinkLocation(entity)'
+					}
+				}
+			}
+			static get contentClass() {
+				return 'link-content-service';
```
jwong-d2l: Onedrive is content-service too?

tinglisD2L:
@jwong-d2l The only two pages that have the exact same layout (large icon, body text, and one action button) are SCORM and OneDrive, so it seems a bit overkill to make that generic.

components/d2l-sequences-content-link-onedrive.html
tinglisD2L: Um I don't know actually. I copied that page from SCORM. @lawong-d2l is going to hook it up to the backend later so there might be some changes then.

jwong-d2l: maybe co-ordinate with him to get the right class then.

jwong-d2l: I called it "link-onedrive"

components/d2l-sequences-content-link-onedrive.html
```diff
@@ -0,0 +1,78 @@
...
+						computed: '_getLinkLocation(entity)'
+					}
+				}
+			}
+			static get contentClass() {
+				return 'link-onedrive';
+			}
+			connectedCallback() {
+				super.connectedCallback();
+				this._language = window.document.getElementsByTagName('html')[0]
```
dbatiste: I don't think this should be needed since our localization lib handles this, as well as other things.

jwong-d2l:
I added the routing logic for you, somehow that was missed.  Should probably update the tests for the router to also test for our new path to onedrive.

components/d2l-sequences-content-link-onedrive.html
jwong-d2l: Okay, we'll look into removing it in another PR.  There are a lot of places we have this.

PR #45, opened 2018-09-28
add end of lesson images

tinglisD2L:


lawong-d2l:
I was unaware we needed to make two new components to show these pictures, rather than just saving them as SVG files in an asset folder and rendering them through another container. Is this the correct pattern?

tinglisD2L:
@lawong-d2l I did that at first, and while it worked for `polymer serve`, it didn't work with `npm run build` because the latter was looking in the dist folder. I couldn't figure out how to get the path correct. If you can find any examples of how it's done, I'd be happy to try it, but so far all the image assets we're using seem to be wrapped in icon classes.

PR #46, opened 2018-10-01
Tinglis/use d2l navigation icon

tinglisD2L:


bower.json
```diff
@@ -16,21 +16,18 @@
   ],
   "dependencies": {
     "polymer": "Polymer/polymer#^2.6.0",
-    "d2l-icons": "#4.14.2",
```
jwong-d2l: yessssss

components/d2l-sequences-content-eol-main.html
```diff
@@ -3,6 +3,8 @@
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../d2l-polymer-siren-behaviors/store/entity-behavior.html">
 <link rel="import" href="../localize-behavior.html">
+<link rel="import" href="end-of-lesson-complete-image.html">
+<link rel="import" href="end-of-lesson-incomplete-image.html">
```
jwong-d2l: Was there any thought around combining the two icons into one component and having it be switchable with an attribute (kind of like the notification button?)

components/d2l-sequences-iterator.html
```diff
@@ -1,134 +1,128 @@
...
-        <d2l-icon-button
-            id="iteratorButton"
-            disabled="[[disabled]]"
-            aria-disabled$="[[disabled]]"
-            icon="[[icon]]"
-            type="button"
-            on-click="_onClick"
-        >[[localize(iterateTo)]]</d2l-icon-button>
-    </template>
+	<template>
```
jwong-d2l: was this the autoformat?

tinglisD2L: I've been slowly converting whitespaces to tabs in files as I encounter them.

components/d2l-sequences-content-eol-main.html
tinglisD2L: Could be a nice code cleanup thing to do, but no plans as of yet since it's not really used anywhere else.

jwong-d2l: I was thinking that the two components will never be separate, so why not just use the same component.  Can you make a task for that, and I'll approve

tinglisD2L: @jwong-d2l Sure but can you move/split US99845 in Rally to put into the current sprint so I can add the task? I don't know how to do that. Thanks

jwong-d2l: It's not technically part of that story since the AC didn't specify an image.  We should have made a separate story or task for this whole change actually.  For now, just make a new story and put it in the backlog.

tinglisD2L: oh sure

PR #49, opened 2018-10-04
Lint

benyamin-noori:
Fix lint problems. 
Add `npm run test` to travis CI. 
Fix tests that were failing.


components/d2l-sequences-content-link.html
```diff
@@ -44,26 +44,26 @@
 						type: String,
 						computed: '_getLinkLocation(entity)'
 					}
-				}
+				};
 			}
 			static get observers() {
-				return ['_render(entity)']
+				return ['_render()'];
```
jwong-d2l: what happens if we take out the thing that is being watched? @jbydeley-d2l

locales.json
```diff
@@ -19,10 +19,13 @@
 		"activitiesFinishedGreatJob": "You have completed all of the activities in this lesson. Great job!",
 		"imDone": "I'm Done",
 		"niceWork": "Nice work so far!",
-		"youMayHave": "You may have",
+		"you": "You",
```
gravindran-d2l: Could you verify this one -- if master had "youMayHave", you might be overwriting something.

components/d2l-sequences-content-link.html
jwong-d2l: aside from this one thing, I think it looks good!

locales.json
benyamin-noori: Actually I have no idea how this change happened. It looks like it should be "youMayHave" though. I'll change it back.

jwong-d2l: these are in master, so maybe it's just the merge showing a false alarm? https://github.com/BrightspaceHypermediaComponents/sequences/blob/master/locales.json

jwong-d2l: oh wait, the "you" one is different.

benyamin-noori: I talked to @tinglisD2L and she says it should be "you" and not "youMayHave". So what we have here is the correct version.

gravindran-d2l: Excellent.  Code review FTW!

components/d2l-sequences-content-link.html
jbydeley-d2l: Polymer uses that to know when to trigger the `_render` function. We need to have `entity` in there or it won't fire when `entity` changes.

benyamin-noori: Added a comment to ignore no-unused-vars for this file specifically. I thought we'd still want to have this rule for the rest of the project.

PR #62, opened 2018-10-18
Spinner covers iframe when token changes but href stays the same

jbydeley-d2l:
Fix spinner covering iframe when same href passed in
Fix lint issue of not including correct html import for one drive


components/d2l-sequences-content-link-onedrive.html
```diff
@@ -1,4 +1,4 @@
-<link rel="import" href="../../d2l-content-icons/d2l-content-icons.html">
+<link rel="import" href="../../d2l-content-icons/d2l-open-one-drive-icon.html">
```
tinglisD2L: oh good catch! I think that was my bad.

PR #63, opened 2018-10-24
Use embed links if available

benyamin-noori:
getLinkLocation function now first checks to see if an embed link is present in the entity and if so, returns that link.

### DIFF
--- a/locales.json
+++ b/locales.json
@@ -15,7 +15,7 @@
 		"goToActivity": "Go to Activity",
 		"openNew": "Open in a new window",
 		"cannotBeRendered": "This content item cannot be rendered.",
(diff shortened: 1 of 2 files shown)


locales.json
```diff
@@ -15,7 +15,7 @@
 		"goToActivity": "Go to Activity",
 		"openNew": "Open in a new window",
 		"cannotBeRendered": "This content item cannot be rendered.",
-		"congratulations": "Congratulations",
+		"congratulations": "Congratulations!",
```
jwong-d2l: was this related?

benyamin-noori: No, it's from an earlier change but when you build it you get this change.

PR #64, opened 2018-10-25
Fix embed link location getter

benyamin-noori:
Instead of calling getLinkByRel we should be calling getLinkByClass to get the embed link.


benyamin-noori:
My bad. I thought 'embed' was in the rel list similar to 'about', but it's in the class. Here's the corresponding change in the LMS: 
https://git.dev.d2l/projects/CORE/repos/le/commits/8e0cb9017441d69f7c5847ecbc3e8324521d353c#content/D2L.LE.Content/Extensibility/HM/Content/Serializers/Siren/SirenContentLinkSerializer.cs

jwong-d2l:
whoops totally missed that.. haha

PR #71, opened 2018-11-27
Send click-back event when Done button clicked

jbydeley-d2l:
Part 2

We'll send a `back-click` event from the EoL `I'm Done` button to signal the viewer to navigate to the `returnUrl`. We don't have access to it on the EoL page anymore since everything goes through the router and the router only passes href and token.

### DIFF
--- a/mixins/d2l-sequences-return-mixin.html
+++ b/mixins/d2l-sequences-return-mixin.html
@@ -3,7 +3,7 @@
     	function ReturnMixin(classes) {
     		return class extends Polymer.mixinBehaviors(classes, Polymer.Element) {
     			_onClickBack() {
-    				const event = new CustomEvent('click-back', {
+				const event = new CustomEvent('sequences-return-mixin-click-back', {
     					composed: true,
     					bubbles: true
     				});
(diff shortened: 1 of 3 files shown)


mixins/d2l-sequences-return-mixin.html
```diff
@@ -0,0 +1,22 @@
+<script>
+    (function() {
+    	function ReturnMixin(classes) {
+    		return class extends Polymer.mixinBehaviors(classes, Polymer.Element) {
+    			_onClickBack() {
+    				const event = new CustomEvent('click-back', {
```
dbatiste: Nit: a general best practice for custom events is to qualify them.  In this case there isn't likely to be a collision, but if we prefix it with the component/mixin name them there's no questioning where it's coming from (when reading the listener code).

jbydeley-d2l: Would you use `return-mixin-click-back` or `d2l-sequences-return-mixin-click-back`? I'm happy to change it.

jbydeley-d2l: I updated it to `sequences-return-mixin-click-back`.

PR #72, opened 2018-11-30
Join two lang terms into one

jbydeley-d2l:


lang/en.json
```diff
@@ -17,7 +17,7 @@
 	"imDone": "I'm Done",
 	"niceWork": "Nice work so far!",
 	"you": "You",
```
benyamin-noori: I think we should remove `you` now that's no longer being used anywhere.

jwong-d2l: good catch

PR #79, opened 2018-12-20
Calling setDashboardViewState for modules and topics

lawong-d2l:
When they get viewed in the content router call setDashboardViewState

### DIFF
--- a/mixins/d2l-sequences-completion-tracking-mixin.html
+++ b/mixins/d2l-sequences-completion-tracking-mixin.html
@@ -28,6 +28,25 @@
 					};
 				}
 
+				topicSetDashboardViewState() {
+					if (!this.entity) {
+						return;
+					}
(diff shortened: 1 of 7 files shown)


mixins/d2l-sequences-completion-tracking-mixin.html
```diff
@@ -28,6 +28,25 @@
 					};
 				}
 
+				topicSetDashboardViewState() {
+					if (!this.entity) {
+						return;
+					}
+					return new Promise((resolve, reject) => {
```
benyamin-noori: It looks like we have this code in two places. Is there an easy way for us to not have two instances of it?

lawong-d2l: @benyamin-noori fixed, good suggestion. However I want to keep module by itself, since the completion tracking mixin is very topic specific.

PR #81, opened 2019-01-11
Add new property error, show uknown page when error

benyamin-noori:
Add a new property 'error' to the element and use the render function to observer changes. When error occurs, we replace the current element in the page with a new component,  `D2LSequenceContentError`. If any additional dialogs are open, they will be closed.

We can use this new element to customize the contents of the page based on the the type of error, etc. 

Important note: If this change is merged, the LMS error page will no longer be shown/available to the user. 

Related, LMS-side PRs: 

https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/2820/overview
https://git.dev.d2l/projects/CORE/repos/le/pull-requests/13645/overview

### DIFF
--- a/components/d2l-sequences-content-error.html
+++ b/components/d2l-sequences-content-error.html
@@ -1,0 +1,16 @@
+<link rel="import" href="../polymer/polymer.html">
+<dom-module id="d2l-sequences-content-error">
+	<template>
+		<style>
+		</style>
+		[[localize('cannotBeRendered')]]
+	</template>
+	<script>
+		class D2LSequencesContentError extends Polymer.Element {
+			static get is() {
+				return 'd2l-sequences-content-error';
+			}
+		}
+		customElements.define(D2LSequencesContentError.is, D2LSequencesContentError);
+	</script>
+</dom-module>
--- a/components/d2l-sequences-content-error.html
+++ b/components/d2l-sequences-content-error.html
@@ -1,3 +1,4 @@
+<link rel="import" href="../localize-behavior.html">
 <dom-module id="d2l-sequences-content-error">
 	<template>
 		<style>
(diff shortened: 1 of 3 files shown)


components/d2l-sequences-content-error.html
```diff
@@ -0,0 +1,20 @@
+<link rel="import" href="../localize-behavior.html">
+<dom-module id="d2l-sequences-content-error">
+	<template>
+		<style>
+		</style>
+		[[localize('cannotBeRendered')]]
```
lawong-d2l: Do we need any designs for what these error pages should look like?

benyamin-noori: I don't think we do. I looked through the UI/UX design in invision and zeppelin but didn't find anything. 

Should we ask for a design from Nick?

jwong-d2l: I think that's sufficient for now.

mixins/d2l-sequences-router-mixin.html
```diff
@@ -23,36 +24,52 @@
...
-					return ['_render(entity)'];
+					return ['_render(entity, error)'];
 				}
 
 				connectedCallback() {
 					super.connectedCallback();
 					this._hrefUpdated = this.addEventListener('hrefUpdated', ({detail}) => {
 						this.href = detail.href;
 					});
+					this._D2lErrorListener = window.addEventListener('message', (event) => {
```
jwong-d2l: should this be D2LErrorListener?

benyamin-noori: Actually, I capitalized it like this this according to something I read in docs, but I think that was for C sharp. I'll fix this one.

PR #85, opened 2019-01-25
Fix localization issues

jbydeley:
Change node scripts to use `require` instead of `import`. `import` isn't natively supported by node.
Update gulp to 4.0.0
Update plop handlebars template to generate polymer 3 code instead of polymer 2

### DIFF
--- a/plop/localize-behavior.hbs
+++ b/plop/localize-behavior.hbs
@@ -1,30 +1,31 @@
-<link rel="import" href="../polymer/polymer.html">
-<link rel="import" href="../d2l-localize-behavior/d2l-localize-behavior.html">
-<script>
-    /* eslint-disable */
-    window.D2L = window.D2L || {};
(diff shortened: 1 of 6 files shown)


jwong-d2l:
do we need a minor version bump too?

plop/localize-behavior.hbs
```diff
@@ -1,30 +1,31 @@
...
+import '@polymer/polymer/polymer-legacy.js';
+import 'd2l-localize-behavior/d2l-localize-behavior.js';
+
+/* eslint-disable */
+window.D2L = window.D2L || {};
+window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+window.D2L.PolymerBehaviors.Sequences = window.D2L.PolymerBehaviors.Sequences || {};
+
+/**
+ * THIS FILE IS GENERATED. RUN `npm run locales` TO REGENERATE.
```
jwong-d2l: oops, this needs to be updated to `npm run build:lang`

PR #86, opened 2019-01-31
Using the redirect to iterate

lawong-d2l:
https://github.com/Brightspace/d2l-sequence-viewer/pull/142


components/d2l-sequences-iterator.js
```diff
@@ -9,7 +9,7 @@ import { PolymerElement } from '@polymer/polymer/polymer-element.js';
 	@extends D2L.PolymerBehaviors.Sequences.LocalizeBehavior
 	@demo demos/index.html
 */
-class D2LSequencesIterator extends mixinBehaviors([
+export class D2LSequencesIterator extends mixinBehaviors([
```
jbydeley-d2l: Why are you exporting this? I don't see you importing it anywhere... or is that in a different review?

PR #88, opened 2019-02-06
Use proper ES6 module imports in tests

jbydeley:
Use Element.is instead of hard coding element tags
Update polymer-siren-test-helpers to version 2 to support ES6 modules


jbydeley:
This account has no merge privileges so you'll have to merge it for me when you are happy with it.

PR #89, opened 2019-02-06
use pdf link, if available, for documents

jwong-d2l:


components/d2l-sequences-content-file-html.js
```diff
@@ -44,7 +44,7 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 		try {
 			const fileActivity = entity.getSubEntityByClass('file-activity');
 			const file = fileActivity.getSubEntityByClass('file');
-			const link = file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			const link = file.getLinkByClass('pdf') || file.getLinkByRel('alternate');
```
benyamin-noori: Don't we want the embed link as a final option? Don't we want to use that at all anymore?

jwong-d2l: talked to you offline, there's no embed link for file-html types, only for link types of content.

PR #90, opened 2019-02-13
add wordprocesing to the known types list

jwong-d2l:


jwong-d2l:
This is just a hacky workaround for now to test the pdf links working.  We need a real fix here, that covers all the different possible types.

PR #91, opened 2019-02-15
use embed-asv link if it's available

benyamin-noori:


jwong-d2l:
might want to discard changes to package-lock

PR #95, opened 2019-02-21
Update package.json

jwong-d2l:


jwong-d2l:
I did this as a test so that we're not double travising.  looks like it worked.

PR #98, opened 2019-03-11
US104610/multipage topic navigation in asv

asilver-d2l:
Allows iterators to navigate through multi-page content. Variables that determine if it is multi-page content gets reset on unload of html content.

### DIFF
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -42,8 +42,9 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 	}
 
 	disconnectedCallback() {
+		super.disconnectedCallback();
 		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
-		window.postMessage(JSON.stringify({ handler: 'd2l.nav.reset' }), '*');
+		window.dispatchEvent(new CustomEvent('d2l-nav-request-customizations', {detail: JSON.stringify({ handler: 'd2l.nav.reset' })}));
 	}
 
 	_scrollToTop() {
@@ -65,17 +66,12 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 	}
 
 	_navigateMultiPageFile(e) {
-		try {
-			const iframe = this.shadowRoot.querySelector('iframe');
-			if (iframe) {
-				iframe.contentWindow.postMessage(JSON.stringify({
-					handler: 'd2l.nav.client',
-					action: e.detail
-				}), '*');
-			}
-		}
-		catch (e) {
-			return '';
+		const iframe = this.shadowRoot.querySelector('iframe');
+		if (iframe && iframe.contentWindow) {
+			iframe.contentWindow.dispatchEvent(new CustomEvent('d2l-sequences-multipage-nav', { detail: JSON.stringify({
+				handler: 'd2l.nav.client',
+				action: e.detail
+			})}));
 		}
 	}
 }
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -43,7 +43,7 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 
 	disconnectedCallback() {
 		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
-		window.postMessage( JSON.stringify({ handler: "d2l.nav.reset" }) , '*' );
+		window.postMessage(JSON.stringify({ handler: 'd2l.nav.reset' }), '*');
 	}
 
 	_scrollToTop() {
@@ -71,7 +71,7 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 				iframe.contentWindow.postMessage(JSON.stringify({
 					handler: 'd2l.nav.client',
 					action: e.detail
-				}), "*");
+				}), '*');
 			}
 		}
 		catch (e) {
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -43,6 +43,7 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 
 	disconnectedCallback() {
 		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
+		window.postMessage( JSON.stringify({ handler: "d2l.nav.reset" }) , '*' );
 	}
 
 	_scrollToTop() {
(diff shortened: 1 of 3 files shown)


components/d2l-sequences-content-file-html.js
```diff
@@ -36,6 +36,16 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
...
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this.multiPageNavListener = window.addEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFile.bind(this));
+	}
+
+	disconnectedCallback() {
+		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
+		window.postMessage( JSON.stringify({ handler: "d2l.nav.reset" }) , '*' );
```
benyamin-noori: single quotes for d2l.nav.reset?

components/d2l-sequences-content-file-html.js
```diff
@@ -53,5 +63,20 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
...
+		try {
+			const iframe = this.shadowRoot.querySelector('iframe');
+			if (iframe) {
+				iframe.contentWindow.postMessage(JSON.stringify({
+					handler: 'd2l.nav.client',
+					action: e.detail
+				}), '*');
+			}
+		}
+		catch (e) {
```
jwong-d2l: what exceptions can be thrown?

components/d2l-sequences-iterator.js
```diff
@@ -68,6 +68,15 @@ export class D2LSequencesIterator extends mixinBehaviors([
...
 				type: String,
 				computed: 'getIterateTo()'
+			},
+			isMultiPage: {
+				type: Boolean
+			},
+			multiPageHastNext: {
+				type: Boolean
+			},
+			multiPageHastPrev: {
```
jwong-d2l: typo... unless you're doing old english.

jwong-d2l: then I think it should be hath?

components/d2l-sequences-iterator.js
```diff
@@ -77,9 +86,38 @@ export class D2LSequencesIterator extends mixinBehaviors([
...
+
+	disconnectedCallback() {
+		window.removeEventListener('message', this._setUpMultiPageTopic.bind(this));
+	}
+
+	_setUpMultiPageTopic(e) {
+		const data = JSON.parse(e.data);
+		if (data && data.handler === 'd2l.nav.customize') {
+			this.isMultiPage = true;
+			if (data.hasNext) {
```
jwong-d2l: does ```this.multiPageHasNext = Boolean(data.hasNext)``` work?

jwong-d2l:
definitely want to add some tests  - there are some easy ones that can include the new conditional logic with multipage topics.

components/d2l-sequences-iterator.js
```diff
@@ -77,9 +86,38 @@ export class D2LSequencesIterator extends mixinBehaviors([
...
 		];
 	}
+
 	connectedCallback() {
 		super.connectedCallback();
+		window.addEventListener('message', this._setUpMultiPageTopic.bind(this));
 	}
+
+	disconnectedCallback() {
+		window.removeEventListener('message', this._setUpMultiPageTopic.bind(this));
```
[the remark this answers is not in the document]
jwong-d2l: also, the event name is 'message'?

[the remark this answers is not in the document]
jwong-d2l: oh, I see, it's coming from postMessage.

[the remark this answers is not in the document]
jwong-d2l: https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Security_concerns

components/d2l-sequences-content-file-html.js
asilver-d2l: There's a bunch of `Cannot read property 'postMessage' of null` errors that pop up on the console without the try-catch.

jwong-d2l: maybe you should switch it to ```if(iframe && iframe.contentWindow)``` ?

components/d2l-sequences-iterator.js
asilver-d2l: Yes, that seems to work.

components/d2l-sequences-content-file-html.js
asilver-d2l: That seems to stop throwing the error without using try-catch.

components/d2l-sequences-iterator.js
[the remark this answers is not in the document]
asilver-d2l: Would this be better as a custom event? `d2l-nav.js` has a `postMessage()` event and a `message` event listener. Should those also be changed? Are we expecting the html iframe to be from the same origin all of the time?

components/d2l-sequences-iterator.js
```diff
@@ -104,5 +150,21 @@ export class D2LSequencesIterator extends mixinBehaviors([
...
+			return 'next';
+		} else if (this.previous) {
+			return 'prev';
+		}
+		return '';
+	}
+
+	_isMultiPageNavigation() {
+		if (this.isMultiPage) {
+			if (this.next && this.multiPageHastNext) return true;
```
jwong-d2l: So we require a next and a previous to be a multipage topic?

asilver-d2l: I don't think so. I think this is if you say, press the next iterator button and the multi-page topic has a "next page", then it will do the multi-page navigation stuff. If you press next and there is no next page in the multi-page topic, you will do the regular navigation.

components/d2l-sequences-iterator.js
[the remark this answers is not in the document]
jwong-d2l: Yeah, maybe. I think the iframe event can bubble up to the parent frame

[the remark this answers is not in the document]
asilver-d2l: What do you mean exactly by "bubble up"?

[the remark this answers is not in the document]
jwong-d2l: https://developer.mozilla.org/en-US/docs/Web/API/Event/bubbles

jwong-d2l:
let me know when you're done with those changes and I'll take another look

components/d2l-sequences-content-file-html.js
```diff
@@ -36,6 +36,16 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 		};
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this.multiPageNavListener = window.addEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFile.bind(this));
+	}
+
+	disconnectedCallback() {
+		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
```
jwong-d2l: need ```super.disconnectedCallback();```

components/d2l-sequences-content-file-html.js
```diff
@@ -36,6 +36,17 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 		};
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this.multiPageNavListener = window.addEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFile.bind(this));
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
```
jwong-d2l: didn't refresh... whoops.

PR #99, opened 2019-03-12
Remove references to set-dashboard-state-view siren action

benyamin-noori:
This siren action no longer exists in the response. The logic behind this action has been added to set-last-viewed-content-object.

### DIFF
--- a/test/d2l-sequences-automatic-completion-tracking-mixin.html
+++ b/test/d2l-sequences-automatic-completion-tracking-mixin.html
@@ -44,13 +44,11 @@ describe('d2l-sequences-automatic-completion-tracking-mixin', () => {
 		element = fixture('MixinFixture');
 	});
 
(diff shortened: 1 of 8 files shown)


test/d2l-sequences-automatic-completion-tracking-mixin.html
```diff
@@ -44,13 +44,11 @@
 		element = fixture('MixinFixture');
 	});
 
-	it('startCompletion, topicSetDashboardViewState and finishCompletion must be called when element is updated', async() => {
+	it('startCompletion and finishCompletion must be called when element is updated', async() => {
```
lawong-d2l: Do we have tests here to cover the setting of last viewed object being called?

benyamin-noori: No, but I will add them. They'll go into the sequence viewer and not this repo.

PR #100, opened 2019-03-14
Add tests for multi-page content

asilver-d2l:
- Adds tests for multi-page content
- Adds multi-page content demo page


asilver-d2l:
I was using the other json for tests so that I could load up the second page to test if clicking the previous iterator would work and the values of the variables on the iterators. Would it make more sense if I just clumped up the tests all in one then?

asilver-d2l:
Or should the tests stay split, but use the iterators to go to the second page to do the tests that I want to do on the second page?

benyamin-noori:
> Or should the tests stay split, but use the iterators to go to the second page to do the tests that I want to do on the second page?

I would say a "real" scenario is better (i.e. using the iterators to navigate and see if the properties are set correctly).

PR #101, opened 2019-03-25
re-introduce old scorm support

jwong-d2l:


lawong-d2l:
I worry, that this might not be extensible for future releases of scorm, but as it stands these are the versions we are aware of

PR #104, opened 2019-03-28
add missing const import

jwong-d2l:


lawong-d2l:
This makes me super concerned about why we haven't seen this come up, and how many more places are we missing a reference.,

jwong-d2l:
> This makes me super concerned about why we haven't seen this come up, and how many more places are we missing a reference.,

polymer html imports cascaded our imports for us, but now real js imports are screwing us.

PR #107, opened 2019-04-04
added open link in new tab for sequence

iFun:


components/d2l-sequences-content-link-new-tab.js
```diff
@@ -0,0 +1,89 @@
...
+			}
+
+		</style>
+		<div class="content-link-new-tab-container">
+			<img id="linkImage" src="../img/weblink-jump-image.svg"></img>
+			<p>
+				[[localize('openInNewTab')]]
+			</p>
+			<d2l-button primary="" on-click="_onclick">
+				[[localize('openNew')]]
```
benyamin-noori: When you have langterms, don't we need a `connectedCallback` call to load the localization resources (e.g. langterms)?

components/d2l-sequences-content-link-new-tab.js
```diff
@@ -0,0 +1,89 @@
...
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		this.finishCompletion();
+	}
+
+	_scrollToTop() {
+		window.top.scrollTo(0, 0);
+	}
+
+	_getLinkLocation(entity) {
```
benyamin-noori: Another way to set this up would be to have a property in the component and set `computed` to `_getLinkLocation` and when you need it in `_onClick` you can just say `this.propName`. 

I'm saying this because that's how it's been implemented in other components but I'm not sure if there are any significant advantage/disadvantages to it.

components/d2l-sequences-content-link-new-tab.js
iFun: I copied from mixcontent, the langterms worked after I ran npm run build:lang

PR #110, opened 2019-04-09
move svg

iFun:


m6jones:
Why the move?

iFun:
I was having trouble to get the svg src url works in the sequence viewer.

PR #111, opened 2019-04-15
Added video support to sequnece

iFun:
uses d2l-video/d2l-audio to support viewing video/audio inside sequence


jwong-d2l:
d2l-video is not a public repo... as this is a public repo, you'll need to either add a key or make d2l-video public.

iFun:
> d2l-video is not a public repo... as this is a public repo, you'll need to either add a key or make d2l-video public.

I see I will talk to the owner

.travis.yml
```diff
@@ -1,8 +1,11 @@
...
-addons:
-  chrome: stable
-script:
-- npm run lint
-- npm run test
\ No newline at end of file
+language: node_js
+node_js: 9
+before_install:
+- curl -ks -u$ARTIFACTORY_USER:$ARTIFACTORY_PASS "https://d2lartifacts.artifactoryonline.com/d2lartifacts/api/npm/npm-local/auth/d2l" -o .npmrc
```
jwong-d2l: I think the 
```suggestion
- curl -ks -u$ARTIFACTORY_USER:$ARTIFACTORY_PASS "https://d2lartifacts.jfrog.io/d2lartifacts/api/npm/npm-local/auth/d2l" -o .npmrc
```
I think the artifactoryonline.com domain has been deprecated, and should be jfrog.io.

components/d2l-sequences-content-video.js
```diff
@@ -0,0 +1,68 @@
...
+		};
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+	}
+
+	disconnectedCallback() {
+		super.disconnectedCallback();
+		window.dispatchEvent(new CustomEvent('d2l-nav-request-customizations', { detail: JSON.stringify({ handler: 'd2l.nav.reset' }) }));
```
jwong-d2l: I *think* this is only needed for multipage html topics. @benyamin-noori can you comment?

iFun: i will delete this

benyamin-noori: Yes. This line is only used for multi-page topics.

jwong-d2l:
looks good to me, after you remove those lines we talked about.

iFun:
@jwong-d2l could you approve the PR?

components/d2l-sequences-content-video.js
jwong-d2l: uh, these lines are still here.

iFun: removed

.travis.yml
```diff
@@ -1,8 +1,11 @@
...
-addons:
-  chrome: stable
-script:
-- npm run lint
-- npm run test
\ No newline at end of file
+language: node_js
+node_js: 9
+before_install:
+- curl -ks -u$ARTIFACTORY_USER:$ARTIFACTORY_PASS "https://d2lartifacts.jfrog.io/d2lartifacts/api/npm/npm-local/auth/d2l" -o .npmrc
```
m6jones: So adding the private repo has broken lots of builds and this will probably explode in BSI. We may want to think about finding a new solution. You mentioned you talked to the owners of d2l-audio and d2l-video. Why do they need to remain private?

dbatiste: I encountered these during the P3 upgrade as well, and I always seem to stumble when running into it.  I don't see a reason for these to be private, but would be good to know from SMEs.

iFun: I talked to Rhys Causey who is the owner of these two repos He suggested me it's probably easier to just authenticate with Artifactory instead of making it public since there is some private npm repo created by him inside the d2l-video which also needs to be public.

PR #112, opened 2019-04-16
added Google Drive support

iFun:


components/d2l-sequences-content-link-googledrive.js
```diff
@@ -0,0 +1,51 @@
...
+	}
+	static get properties() {
+		return {
+			href: {
+				type: String,
+				reflectToAttribute: true,
+				notify: true,
+				observer: '_scrollToTop'
+			},
+			previousHref: {
```
mpharoah-d2l: Why is this here?

components/d2l-sequences-content-router.js
```diff
@@ -95,8 +96,11 @@ function getLinkEntityType(linkActivity) {
 	const link = linkActivity.getLinkByRel('about');
 	const openInNewTab = linkActivity.hasClass('open-in-new-tab');
 
-	if (link && link.href.startsWith(window.location.protocol)) {
+	if (link && link.href.includes('docs.google')) {
```
mpharoah-d2l: This will miss some links formats and can have false positives. Here's a better check we use elsewhere:
```
const googleDriveRegex = /^https?:\/\/(?:docs|drive)\.google\.com(?:\/a\/[^\/]+)?\/(?!forms\/)[^\/]+\/d\/([^\/\?#]{2,})(?:$|[\/\?#])/;
const alternateLinkFormatRegex = /^https?:\/\/(?:docs|drive)\.google\.com\/open\?id=([^&#]+)(?:$|[&#])/;
const isGoogleDriveLink = googleDriveRegex.exec( url ) || alternateLinkFormatRegex.exec( url );
```

jwong-d2l: any chance we could put this regex in the HM endpoint and add a class instead that marks it as google docs?

mpharoah-d2l: Wish granted
https://git.dev.d2l/projects/CORE/repos/le/pull-requests/15002/overview

components/d2l-sequences-content-link-googledrive.js
iFun: I will remove it

PR #117, opened 2019-04-24
Added video and audio support

iFun:
d2l-video and d2l-audio are public repos now. so there shouldn't be a problem in BSI


iFun:
> Changes look good, but some of the tests are failing.

trying to figure out why, will fix before merge

PR #131, opened 2019-05-28
fix event listener references

apalaniuk:
Noticed a few things re: event listeners while poking around.

### DIFF
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -36,9 +36,14 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 		};
 	}
 
+	constructor() {
+		super();
+		this.multiPageNavListener = this._navigateMultiPageFile.bind(this);
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
-		this.multiPageNavListener = window.addEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFile.bind(this));
+		window.addEventListener('d2l-sequence-viewer-multipage-navigation', this.multiPageNavListener);
 	}
 
 	disconnectedCallback() {
--- a/components/d2l-sequences-iterator.js
+++ b/components/d2l-sequences-iterator.js
@@ -87,14 +87,19 @@ export class D2LSequencesIterator extends mixinBehaviors([
 		];
 	}
 
+	constructor() {
+		super();
+		this._setUpMultiPageTopicBound = this._setUpMultiPageTopic.bind(this);
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
-		window.addEventListener('message', this._setUpMultiPageTopic.bind(this));
+		window.addEventListener('message', this._setUpMultiPageTopicBound);
 	}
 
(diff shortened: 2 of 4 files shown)


components/d2l-sequences-content-file-html.js
```diff
@@ -36,9 +36,14 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
...
 	}
 
+	constructor() {
+		super();
+		this.multiPageNavListener = this._navigateMultiPageFile.bind(this);
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
-		this.multiPageNavListener = window.addEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFile.bind(this));
```
apalaniuk: `addEventListener` [returns undefined](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Return_value), so capture the bound version of the function in the constructor (since `connectedCallback` may be called multiple times) and use that for adding/removing the event listener

components/d2l-sequences-iterator.js
```diff
@@ -87,14 +87,19 @@ export class D2LSequencesIterator extends mixinBehaviors([
...
 	}
 
+	constructor() {
+		super();
+		this._setUpMultiPageTopicBound = this._setUpMultiPageTopic.bind(this);
+	}
+
 	connectedCallback() {
 		super.connectedCallback();
-		window.addEventListener('message', this._setUpMultiPageTopic.bind(this));
```
apalaniuk: `bind` returns a new function, so the function passed in to `addEventListener` is actually different than the one passed to `removeEventListener` - needs to be captures and the same instance passed to both.

mixins/d2l-sequences-router-mixin.js
```diff
@@ -37,20 +37,24 @@ function RouterMixin(getEntityType) {
...
+		connectedCallback() {
+			super.connectedCallback();
+			window.addEventListener('message', this._D2LErrorListener);
 		}
 
 		disconnectedCallback() {
 			super.disconnectedCallback();
-			this.removeEventListener('hrefUpdated', this._hrefUpdated);
-			this.removeEventListener('message', this._D2LErrorListener);
+			window.removeEventListener('message', this._D2LErrorListener);
```
apalaniuk: was being added to `window` in connected callback but removed from `this` in disconnected

mixins/d2l-sequences-router-mixin.js
```diff
@@ -37,20 +37,24 @@ function RouterMixin(getEntityType) {
...
+		}
+
+		connectedCallback() {
+			super.connectedCallback();
+			window.addEventListener('message', this._D2LErrorListener);
 		}
 
 		disconnectedCallback() {
 			super.disconnectedCallback();
-			this.removeEventListener('hrefUpdated', this._hrefUpdated);
```
apalaniuk: event listeners added to the custom element or children don't need to be explicitly removed, but some like seeing symmetry in adds/removes (though adding it to the `listeners` map on the element could make it cleaner still)

jwong-d2l:
looks like these changes have been brought in already in a different commit.  Thanks!

PR #139, opened 2019-06-11
Sequence module list to use SDK

m6jones:
Requires: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/41

Convert the `d2l-sequence-module-list` to use the new sdk!

### DIFF
--- a/components/d2l-sequences-module-list.js
+++ b/components/d2l-sequences-module-list.js
@@ -145,7 +145,7 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 			},
 			_modulesBySequence: {
 				type: Array,
-				value: function() { return {}; }
+				value: function() { return []; }
 			},
 			_modules: {
 				type: Array,
--- a/components/d2l-sequences-module-list.js
+++ b/components/d2l-sequences-module-list.js
@@ -143,10 +143,6 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 				type: String,
 				value: 'd2l-tier1:arrow-collapse'
 			},
-			_modulesBySequence: {
-				type: Array,
-				value: function() { return []; }
-			},
 			_modules: {
 				type: Array,
 				value: function() { return []; }
@@ -191,11 +187,12 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 	}
 
 	_onSequenceRootChange(sequenceRoot) {
+		const modulesBySequence = [];
 		sequenceRoot.onSubSequencesChange((subSequence) => {
-			this._modulesBySequence[subSequence.index()] = {
+			modulesBySequence[subSequence.index()] = {
 				title: subSequence.title()
 			};
-			this._modules = this._modulesBySequence.filter(element => typeof(element) !== 'undefined');
+			this._modules = modulesBySequence.filter(element => typeof(element) !== 'undefined');
 		});
 	}
 
--- a/components/d2l-sequences-module-list.js
+++ b/components/d2l-sequences-module-list.js
@@ -8,7 +8,6 @@ import 'd2l-icons/d2l-icon.js';
 import 'd2l-colors/d2l-colors.js';
 import 'd2l-icons/tier1-icons.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
-import './d2l-sequences-module-name.js';
 import '../localize-behavior.js';
 
 const behaviors = [
--- a/components/d2l-sequences-module-list.js
+++ b/components/d2l-sequences-module-list.js
@@ -143,6 +143,10 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 				type: String,
 				value: 'd2l-tier1:arrow-collapse'
 			},
+			_modulesBySequence: {
+				type: Array,
+				value: function() { return {}; }
+			},
 			_modules: {
 				type: Array,
 				value: function() { return []; }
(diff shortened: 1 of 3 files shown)


components/d2l-sequences-module-list.js
```diff
@@ -112,7 +113,7 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, PolymerElement) {
 						<template is="dom-repeat" items="[[_modules]]">
 							<li>
 								<d2l-link href="#" on-focus="_onFocus" on-blur="_onBlur">
-									<d2l-sequences-module-name module$="[[localize('module')]]" href="[[item]]" token="[[token]]"></d2l-sequences-module-name>
+									<div module$="[[localize('module')]]">[[item.title]]</div>
```
m6jones: The reason I am not using the `d2l-sequence-module-name` is because it only prints the name. With the SDK I can do this in a much better performant way.

components/d2l-sequences-module-list.js
```diff
@@ -143,17 +144,25 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, PolymerElement) {
...
 				type: Array,
-				value: () => []
+				value: function() { return []; }
 			}
 		};
 	}
 	static get observers() {
 		return [
-			'_handleModuleData(entity)'
+			'_onSequenceRootChange(_entity)'
```
m6jones: `_entity` is defined in `EntityMixin`
Polymer magic allows me to watch for any changes.

components/d2l-sequences-module-list.js
```diff
@@ -182,10 +191,13 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, PolymerElement) {
...
-	_handleModuleData(modules) {
-		this._modules = modules && modules.getSubEntities
-			? modules.getSubEntities('https://api.brightspace.com/rels/sequenced-activity').map(e => e.href)
-			: [];
+	_onSequenceRootChange(sequenceRoot) {
+		sequenceRoot.onSubSequencesChange((subSequence) => {
+			this._modulesBySequence[subSequence.index()] = {
+				title: subSequence.title()
+			};
+			this._modules = Object.keys(this._modulesBySequence).sort((a, b) => a < b).map(index => this._modulesBySequence[index]);
```
m6jones: I need an array to use these values in the `dom-repeat` I also need to reset it so it knows it has been updated. Polymer troubles.

jkehres:
When do you decide to bump the package.json version? Does this change warrant a bump?

m6jones:
I tend to do the npm way. But It is faster if I do it in the PR.

components/d2l-sequences-module-list.js
```diff
@@ -143,17 +144,25 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, PolymerElement) {
...
 				value: 'd2l-tier1:arrow-collapse'
 			},
+			_modulesBySequence: {
+				type: Object,
+				value: function() { return {}; }
+			},
 			_modules: {
 				type: Array,
-				value: () => []
+				value: function() { return []; }
```
jkehres: Isn't this functionally equivalent?

jkehres:
What is the "npm way"?

package.json
```diff
@@ -40,7 +40,7 @@
     "wct-browser-legacy": "^1.0.1",
     "whatwg-fetch": "^3.0.0"
   },
-  "version": "1.2.1",
+  "version": "1.3.1",
```
jkehres: If I'm going to be picky, this should be 1.3.0. A minor version bump resets the patch version.

components/d2l-sequences-module-list.js
m6jones: Polymer doesn't run `() => []` so it fails to become the right value. :(

package.json
m6jones: Previous person didn't update the file. So the last version was `1.3.1`

components/d2l-sequences-module-list.js
jkehres: Should you remove the import on line 11 then?

m6jones: Yup

m6jones:
I forgot, here is the picture 
image

Ignore the checkmarks as they are a couple of branches ahead.

components/d2l-sequences-module-list.js
jkehres: Doesn't `Object.keys()` return an array of strings? The sort you're doing appears to assume it's operating on an array of numbers. Doing a less than comparison on stringified numbers probably works for single digit values but not multi-digit values. Also, `Array.prototype.keys()` is a thing. You could use that instead of `Object.keys()` (e.g. `this._modulesBySequence.keys().sort()...`).

m6jones: Nice catch!

components/d2l-sequences-module-list.js
```diff
@@ -182,10 +190,13 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, PolymerElement) {
 		return cond ? t : f;
 	}
 
-	_handleModuleData(modules) {
-		this._modules = modules && modules.getSubEntities
-			? modules.getSubEntities('https://api.brightspace.com/rels/sequenced-activity').map(e => e.href)
-			: [];
+	_onSequenceRootChange(sequenceRoot) {
+		sequenceRoot.onSubSequencesChange((subSequence) => {
+			this._modulesBySequence[subSequence.index()] = {
```
jkehres: Hmm. If a sequence changes and an item is removed, it won't be removed from this array.

package.json
jkehres: You mean last version was 1.3.0?

jkehres: Or you mean you're setting it to the current/proper version and you're going to use `npm version` later to bump it?

PR #140, opened 2019-06-12
Add link

m6jones:
We want these sub modules to be links to the actual content.
So add the link.

### DIFF
--- a/components/d2l-sequences-module-list.js
+++ b/components/d2l-sequences-module-list.js
@@ -111,7 +111,7 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 					<ol class="d2l-sequences-module-list-list">
 						<template is="dom-repeat" items="[[_modules]]">
 							<li>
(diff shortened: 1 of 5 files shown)


components/d2l-sequences-module-list.js
```diff
@@ -111,7 +111,7 @@ class D2lSequenceModuleList extends mixinBehaviors(behaviors, EntityMixin(Polyme
 					<ol class="d2l-sequences-module-list-list">
 						<template is="dom-repeat" items="[[_modules]]">
 							<li>
-								<d2l-link href="#" on-focus="_onFocus" on-blur="_onBlur">
+								<d2l-link href$="[[item.href]]" on-focus="_onFocus" on-blur="_onBlur">
```
m6jones: `$=` notation means if it item.href is undefined then the attribute won't show up.

jkehres: When would the `href` be undefined? Or are you just being defensive?

PR #155, opened 2019-09-03
Redirect content service embed views if Safari or iOS.

qwangd2l:
This overcomes the problem of Safari not allowing iframe content to set cookies for content service embedded content.

### DIFF
--- a/components/d2l-sequences-content-content-service-link.js
+++ b/components/d2l-sequences-content-content-service-link.js
@@ -1,0 +1,65 @@
+import { D2LSequencesContentLink } from './d2l-sequences-content-link.js';
+
+export class D2LSequencesContentContentServiceLink extends D2LSequencesContentLink {
+	static get is() {
+		return 'd2l-sequences-content-content-service-link';
+	}
+
+	static get properties() {
+		return {
+			redirectCs: Boolean,
+			csRedirectPath: String,
+			noRedirectQueryParamString: String
+		};
+	}
+
+	_redirect() {
+		if (this.redirectCs &&
+			(this._isIOS() || this._isSafari()) &&
+			this.csRedirectPath &&
+			this.noRedirectQueryParamString) {
+			const location = window.location.href;
+			const queryParams = this._parseUrlQueryParameters(location);
+			const redirectAfterCS = `${location}${Object.keys(queryParams).length ? '&' : '?'}${this.noRedirectQueryParamString}`;
+			const finalForwardUrl = `${this.csRedirectPath}/${encodeURIComponent(redirectAfterCS)}`;
+
+			window.location.replace(finalForwardUrl);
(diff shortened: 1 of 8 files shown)


components/d2l-sequences-content-content-service-link.js
```diff
@@ -0,0 +1,65 @@
...
+
+	static get properties() {
+		return {
+			redirectCs: Boolean,
+			csRedirectPath: String,
+			noRedirectQueryParamString: String
+		};
+	}
+
+	_redirect() {
```
ChrisKraljevicD2L: Not a big fan of having the code from [`D2L.ContentService`](https://git.dev.d2l/projects/CORE/repos/lms/browse/contentservice/D2L.ContentService.Integrations/Content/AddContent/DialogCall.js?at=7ade8b563c8f121066fdaad9202fc4614b4b2f88) get copy-pasted around. If this is the only place outside of the LMS where it has to exist then I'm okay with it but something tells me that's not the case. Otherwise I'd throw these shared functions (`_redirect`, `_isIOS`, `_isSafari`, etc.) in their own ES6 module, with just `_redirect` `export`ed, that can be `import`ed where it's needed.

It can even live here in `sequences` and other repos can add this as a dependency to `import` it.

package.json
```diff
@@ -41,7 +41,7 @@
     "wct-browser-legacy": "^1.0.1",
     "whatwg-fetch": "^3.0.0"
   },
-  "version": "1.3.13",
+  "version": "1.3.14",
```
ChrisKraljevicD2L: New functionality, new minor version: `1.4.0`

components/d2l-sequences-content-content-service-link.js
qwangd2l: Yeah, I agree it's not ideal. I'm also planning to add similar code to Lessons, but it's slightly different since it's an FRA.
For reference: https://github.com/Brightspace/smart-curriculum/pull/955/files#diff-643d6da4dc219a939344c90b411d30c9L107

Even `_isIOS`, `_isSafari`, and `_parseUrlQueryParameters` are taken from Lessons. 
Aside from Lessons(although slightly different code), this would be the only spot with this code outside of the LMS. This is true until another view emerges.

On the topic of storing them in their own ES6 module:
I suppose, for now it can be extracted as its own utility module and live in Sequences, but I don't think it can be reused in Lessons. It can be reused in future view components I guess?

ChrisKraljevicD2L: That's fair. Perhaps these functions can be made re-usable inside their own repos at least in case they're needed in multiple components?

qwangd2l: Yes, that would be a good step to take.

PR #160, opened 2019-09-18
DE35991 scroll container for content link

ChrisKraljevicD2L:
[Rally story](https://rally1.rallydev.com/#/289692574792/detail/defect/334607127004)

### DIFF
--- a/components/d2l-sequences-content-link.js
+++ b/components/d2l-sequences-content-link.js
@@ -13,6 +13,11 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			.hide {
 				visibility: hidden;
 			}
+			.scroll-container {
+				-webkit-overflow-scrolling: touch;
+				overflow-y: auto;
+				height: 100%;
+			}
 			d2l-loading-spinner {
 				position: fixed;
 				top: 33%;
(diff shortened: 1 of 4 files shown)


components/d2l-sequences-content-link.js
```diff
@@ -23,7 +28,9 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			}
 		</style>
 		<d2l-loading-spinner id="spinner"></d2l-loading-spinner>
-		<iframe id="content" class="hide" frameborder="0" scrolling="auto" src$="[[_linkLocation]]"></iframe>
+		<div class="scroll-container">
+			<iframe id="content" class="hide" frameborder="0" scrolling="auto" src$="[[_linkLocation]]"></iframe>
```
dbatiste: Our standard practice in components is to prefix classes with the component name.  The reason being that in browsers that don't support shadow-DOM, this HTML will be in the same DOM scope as everything else on the page, and is a potential target of any other CSS that's been added to the page.  This is a problem that has come up (in prod) previously.

ChrisKraljevicD2L: oops. I'll fix that and re-release things

PR #161, opened 2019-09-19
DE35991 safe class name

ChrisKraljevicD2L:
[Rally defect](https://rally1.rallydev.com/#/289692574792/detail/defect/334607127004)
Based on Dave Batiste's feedback in #160

### DIFF
--- a/components/d2l-sequences-content-link.js
+++ b/components/d2l-sequences-content-link.js
@@ -13,7 +13,7 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			.hide {
 				visibility: hidden;
 			}
-			.scroll-container {
+			.d2l-sequences-scroll-container {
 				-webkit-overflow-scrolling: touch;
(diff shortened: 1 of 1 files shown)


components/d2l-sequences-content-link.js
```diff
@@ -28,7 +28,7 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			}
 		</style>
 		<d2l-loading-spinner id="spinner"></d2l-loading-spinner>
-		<div class="scroll-container">
+		<div class="d2l-sequences-scroll-container">
 			<iframe id="content" class="hide" frameborder="0" scrolling="auto" src$="[[_linkLocation]]"></iframe>
```
dbatiste: I know `hide` existed previously, but it's super generic and perhaps more at risk than the scroll container one.  Many times, some 3rd party CSS would set `display: none` for such a class, which is seemingly not what we'd want in this case, given that we're setting `visibility`.

PR #164, opened 2019-09-26
DE36119: Module description max width causes scrollbar to be in the middle of the viewer

ianlaidlaw:
- Adjust some styling so that the scroll bar is not in the middle of the module. Now it is on the outside of it.

### DEPENDENCIES:

d2l-sequence-viewer: https://github.com/Brightspace/d2l-sequence-viewer/pull/219

### SCREEN SHOTS:

scrollbar1

scrollbar2


components/d2l-sequences-content-module.js
```diff
@@ -11,11 +12,12 @@ export class D2LSequencesContentModule extends mixinBehaviors([
 		return html`
 		<style include="d2l-typography">
 		:host {
-			max-width: 678px;
-			margin: auto;
 			display: block;
 		}
 
+		.d2l-module-container {
```
ianlaidlaw: There was a div with this class that already wrapped everything but it wasn't being used for anything as far as I could tell.

PR #167, opened 2019-11-13
The beginning of the end

ChrisKraljevicD2L:
Things left to move over from sequence-nav:

- [x] lang
- [x] node dependencies
- [x] components
- [x] demo
- [x] tests (Chris)
- [x] utils
- [x] travis ☠️
- [x] polymer.json
- [x] wct.conf.json

Edit this description and mark yerself as workin on something if you'd like to hop in.
I'll be curating this list as I take a look at what's left to do, and maybe collecting a few check marks together


ChrisKraljevicD2L:
IT'S READY FOR REVIEW

package.json
```diff
@@ -10,14 +10,17 @@
   "scripts": {
     "build:lang": "gulp && npm run gen:localize",
     "build": "npm run build:lang",
+    "gen:localize": "plop localize",
     "lint:fix": "eslint *.js {components,mixins,test,demo}/*.js --fix",
     "lint:js": "eslint . --ext .js,.html test/**/*.js test/**/*.html demo/**/*.js demo/**/*.html",
     "lint": "npm run lint:js",
+    "polymer-lint": "polymer lint -i components/*.js, components/d2l-sequence-navigator/*.js, util/*.js, util/d2l-sequence-navigator/*.js",
```
ianlaidlaw: could you please explain how this is different from the existing `lint` command and if we need to keep both?

ChrisKraljevicD2L: We don't need to keep both, and the difference is this fails for some reason? I'll remove it, we're using regular lint in CI anyways for both nav and sequences.

package.json
```diff
@@ -29,45 +32,58 @@
...
   },
+  "lint-staged": {
+    "*.html": [
+      "eslint",
+      "git add"
+    ]
+  },
   "main": "d2l-sequences.js",
   "dependencies": {
+    "@brightspace-ui/core": "^1",
```
ianlaidlaw: so this might be a big lift, but now that we're bringing in core, are we able to remove some of those older deprecated packages such as `d2l-button` and `d2l-colors`, etc?

ianlaidlaw:
possibly not a big deal but since we're in the process of consolidating our npm packages, it might be worth it to resolve it. I get this warning when I npm install:
`npm WARN sinon-chai@2.14.0 requires a peer of sinon@^1.4.0 || ^2.1.0 || ^3.0.0 || ^4.0.0 but none is installed. You must install peer dependencies yourself.
`

package.json
```diff
@@ -29,45 +32,58 @@
...
     "frau-ci": "^1.40.0",
     "gulp": "^4.0.0",
     "gulp-merge-json": "^1.3.1",
+    "husky": "^0.14.3",
+    "lint-staged": "^4.3.0",
     "plop": "^2.0.0",
-    "polymer-cli": "^1.9.5",
+    "polymer-cli": "^1.9.10",
     "polymer-siren-test-helpers": "^2.0.0",
+    "sinon": "^5.0.7",
```
ianlaidlaw: Could be out of scope, or we could be on this version for a reason - but I noticed that this is 2 major versions behind and there's some cool stuff we could be getting. Might be worth looking into upgrading some of these packages as we move them over. Totally up to you

test/d2l-sequence-navigator/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
ianlaidlaw: jw if this file is needed, this repo already has an eslint config file

ChrisKraljevicD2L: I think this file is meant to override the eslint rules for this test directory, but I already see that the regular ol' `test` folder has an identical `.eslintrc.json`. Imma nuke this.

ianlaidlaw:
Just so I have an idea, could you please explain your reasoning for putting these particular mixins in util -> nav ? instead of the regular mixins folder.

image

ChrisKraljevicD2L:
Hmm, good point. Those mixins were in [sequence nav's `utility` folder](https://github.com/BrightspaceHypermediaComponents/d2l-sequence-navigator/tree/master/utility) originally. Though I don't see why they shouldn't go in `mixins` 🤔 I'll move em around

ChrisKraljevicD2L:
It appears that `sinon-chai` is a dependency of `web-component-tester`, and we already had `sinon@5.0.0` installed, so I think we'd have to downgrade `sinon` to fix that error, since I don't see any new releases of wct.

ianlaidlaw:
> It appears that `sinon-chai` is a dependency of `web-component-tester`, and we already had `sinon@5.0.0` installed, so I think we'd have to downgrade `sinon` to fix that error, since I don't see any new releases of wct.

probably not worth action-ing then I think

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "build:lang": "gulp && npm run gen:localize",
     "build": "npm run build:lang",
+    "deps": "madge ./components ./d2l-sequence-navigator",
```
ianlaidlaw: sweet! so what was the result of this? we don't need to action them yet, I'm just curious if this repo has any problems

ianlaidlaw: also nbd, could you explain naming the script `deps` ?

ianlaidlaw:
LGTM!

PR #172, opened 2020-01-02
DE37195 Ensure iframe unloads before removed from page

ChrisKraljevicD2L:
[Rally Defect](https://rally1.rallydev.com/#/289692574792d/detail/defect/356480369520)

Hooked into `d2l-sequences-router-mixin`'s `_render` function to set an `unload` flag before removing elements from the page, allowing `d2l-sequences-content-link` to unset its `iframe`'s `src` attribute and trigger `before/unload` events.

### DIFF
--- a/components/d2l-sequences-content-link.js
+++ b/components/d2l-sequences-content-link.js
@@ -51,7 +51,7 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			previousHref: {
 				type: String
 			},
-			unload: {
+			isUnloaded: {
 				type: Boolean
 			},
 			_linkLocation: {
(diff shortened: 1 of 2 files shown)


components/d2l-sequences-content-link.js
```diff
@@ -51,14 +51,17 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
 			previousHref: {
 				type: String
 			},
+			unload: {
```
ianlaidlaw: nit: `unload` sounds a bit like an action and not a boolean or a "state".. Maybe consider renaming this to something like `isUnloaded` or something, especially since you already have the function named `unload` ?

components/d2l-sequences-content-link.js
```diff
@@ -51,14 +51,17 @@ export class D2LSequencesContentLink extends D2L.Polymer.Mixins.Sequences.Automa
...
+				type: Boolean
```
ianlaidlaw: nit: should the default `value: false` be included to be explicit? I guess it defaults to false anyways so it's fine

PR #177, opened 2020-01-23
LOR PDF support

ChrisKraljevicD2L:
Tests forthcoming

### DIFF
--- a/components/d2l-sequences-content-file-pdf.js
+++ b/components/d2l-sequences-content-file-pdf.js
@@ -53,9 +53,22 @@ export class D2LSequencesContentFilePdf extends D2L.Polymer.Mixins.Sequences.Aut
 
 	_getFileLocation(entity) {
 		try {
+			linkActivityHref = this._getLinkLocation(entity);
+			if (linkActivityHref) {
+				return linkActivityHref;
+			}
 			const fileActivity = entity.getSubEntityByClass('file-activity');
 			const file = fileActivity.getSubEntityByClass('file');
 			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			return link.href;
+		} catch (e) {
+			return '';
+		}
+	}
+	_getLinkLocation(entity) {
+		try {
+			const linkActivity = entity.getSubEntityByClass('link-activity');
+			const link = linkActivity.getLinkByRel('about');
 			return link.href;
 		} catch (e) {
 			return '';
--- a/components/d2l-sequences-content-file-pdf.js
+++ b/components/d2l-sequences-content-file-pdf.js
@@ -53,7 +53,7 @@ export class D2LSequencesContentFilePdf extends D2L.Polymer.Mixins.Sequences.Aut
 
 	_getFileLocation(entity) {
 		try {
-			linkActivityHref = this._getLinkLocation(entity);
+			const linkActivityHref = this._getLinkLocation(entity);
 			if (linkActivityHref) {
 				return linkActivityHref;
 			}
--- a/components/d2l-sequences-content-router.js
+++ b/components/d2l-sequences-content-router.js
@@ -110,7 +110,7 @@ function getLinkEntityType(linkActivity) {
 	const link = linkActivity.getLinkByRel('about');
 	const openInNewTab = linkActivity.hasClass('open-in-new-tab');
 	if (isPDF(link.href)) {
-		return D2LSequencesContentFilePdf;
+		return D2LSequencesContentFilePdf.is;
 	} else if (link && link.href.startsWith(window.location.protocol)) {
 		return openInNewTab ? D2LSequencesContentLinkNewTab.is : D2LSequencesContentLink.is;
 	} else {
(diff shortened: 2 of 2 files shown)


components/d2l-sequences-content-router.js
```diff
@@ -109,10 +109,18 @@ function getFileEntityType(fileActivity) {
...
 		return D2LSequencesContentLinkMixed.is;
 	}
 }
+
+function isPDF(href) {
+	const ext = '.pdf';
+	const [ hrefNoQSNoHash ] = href.split('?')[0].split('#');
+	const pdfIndex = hrefNoQSNoHash.lastIndexOf(ext);
+	return pdfIndex >= 0 && pdfIndex + ext.length === hrefNoQSNoHash.length;
+}
```
ChrisKraljevicD2L: Please rip this to shreds because I really don't know of a better way to check if a url is a pdf.

components/d2l-sequences-content-file-pdf.js
```diff
@@ -61,6 +65,15 @@ export class D2LSequencesContentFilePdf extends D2L.Polymer.Mixins.Sequences.Aut
 			return '';
 		}
 	}
+	_getLinkLocation(entity) {
+		try {
+			const linkActivity = entity.getSubEntityByClass('link-activity');
+			const link = linkActivity.getLinkByRel('about');
```
ChrisKraljevicD2L: The LOR pdf url will get serialized with an `about` rel.

components/d2l-sequences-content-router.js
```diff
@@ -109,10 +109,18 @@ function getFileEntityType(fileActivity) {
 function getLinkEntityType(linkActivity) {
 	const link = linkActivity.getLinkByRel('about');
 	const openInNewTab = linkActivity.hasClass('open-in-new-tab');
-
-	if (link && link.href.startsWith(window.location.protocol)) {
+	if (isPDF(link.href)) {
```
ChrisKraljevicD2L: Actually, this would be safe to merge since PDFs will render with the normal LOR viewer until the LMS PR gets merged.

components/d2l-sequences-content-router.js
ianlaidlaw: Looks legit to me. I have a question about the last 2 lines, and I'm wondering if it can be simplified. (my approach may have some holes you had already considered)

- After stripping the `?` and `#` parts of the url, can you not just inspect the last bit of the url and determine if it is `.pdf` ?
```
const [ hrefNoQSNoHash ] = href.split('?')[0].split('#');
const myCoolSubString = hrefNoQSNoHash.substring(hrefNoQSNoHash.length - ext.length);

return myCoolSubString === ext;
```

ianlaidlaw: Another super duper minor suggestion would be to rename `hrefNoQSNoHash`. It took a minute for me to realize what `QS` stands for, at first I thought it was `QueStion` for removing the `?` stuff.

IMO long variable names are perfectly fine, especially when they are here to perform some "magic" type stuff which this function is kind of doing. That said I would probably suggest something simple like `hrefBasePath` because that's all this is

ianlaidlaw: One final suggestion is to move this out of this component. Not for reusability / sharability per se, Generally I don't have fond opinions on sharability, but I do think decoupling it from this component makes sense for easier testing, and so this file can remain accurate for what it should be - a `d2l-sequences-content-router` component. we can discuss if people feel differently though.

components/d2l-sequences-content-router.js
```diff
@@ -109,10 +109,18 @@ function getFileEntityType(fileActivity) {
...
+		return D2LSequencesContentFilePdf.is;
+	} else if (link && link.href.startsWith(window.location.protocol)) {
```
ianlaidlaw: I think the first part of this conditional is redundant (the `if link && ..`) part because on line 112 you've already made the assumption that `link` exists by calling `link.href`. Either that line needs this check as well (in which case I would try to look into an earlier exit if `link` doesn't exist... or you can remove it from this line.

ianlaidlaw: Or maybe you need to change this check to `if (link.href && ..`. But unless there's a very real possibility of that property _not_ existing, I wouldn't do that. Redundant null checks are smelly to me

ChrisKraljevicD2L: I think if `link` isn't `null`/`undefined` already then there's a good chance that `href` exists.
The big reason that the `link &&` is on the line below the first part of the conditional is because I completely missed it when I added my PDF check. I'll put in an early exist if `link` doesn't exist

components/d2l-sequences-content-router.js
apalaniuk: I would suggest it's not possible to know if a URL represents any specific file type, and that's what MIME types are for. (Don't fully know the context of this, so maybe a filename check is reasonable, just a thought)

ianlaidlaw: Coming back to this with more context, as mentioned in the LMS pr - I'm not so sure this type of code belongs here. We should look into making the FE dumber and let the api do the heavy lifting.

PR #179, opened 2020-02-10
switching over the labs version of accordion

dlockhart:
The accordion component just got moved over into the "Labs" organization. This updates rubrics to use it, and there's [another PR open to update the other usage in rubrics](https://github.com/Brightspace/d2l-rubric/pull/634). This way, the next person who wants to make a change to accordion won't have to deal with this switchover.


dlockhart:
I wasn't sure if you typically update `analysis.json` and `package-lock.json`? They seem pretty out of date...

ChrisKraljevicD2L:
I wasn't sure if `analysis.json` should be treated more as a build artifact or not, but we definitely haven't updated either in a while.

`package-lock.json` doesn't necessarily need to exist either.

dlockhart:
Yeah like typically `package-lock` is just used in downstream "app projects" like BSI. Want me to remove them both and add to `.gitignore`?

dlockhart:
Updated to remove & ignore lock file and analysis file.

PR #180, opened 2020-02-12
Pbrown/progress end

pbrown-d2l:
One remaining task is to actually run the localization script. NPM currently won't allow me to install some packages as the NPM server is having issues


components/d2l-end-of-lesson-progress.js
```diff
@@ -0,0 +1,41 @@
+import { CompletionStatusMixin } from '../mixins/completion-status-mixin.js';
+import { ASVFocusWithinMixin } from '../mixins/asv-focus-within-mixin.js';
+import '@brightspace-ui/core/components/meter/meter-circle.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+
+class D2LEndOfLessonsProgress extends ASVFocusWithinMixin(CompletionStatusMixin()) {
```
ianlaidlaw: nit:
the filename is `end-of-lesson` but the class is `EndOfLessonS` (with the s)

components/d2l-end-of-lesson-progress.js
```diff
@@ -0,0 +1,41 @@
...
+	static get template() {
+		return html`
+		<style>
+			:host {
+				display: block;
+			}
+
+		</style>
+		<d2l-meter-circle
+						class="d2l-progress"
```
ianlaidlaw: nit: the spacing here is a little off. Looks like lines 16-20 should slide left 3 tabs

components/d2l-end-of-lesson-progress.js
pbrown-d2l: good catch. I will fix this

components/d2l-sequences-content-eol-main.js
```diff
@@ -43,8 +44,9 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 			}
 		</style>
 		<div class="content-eol-main-container">
-			 <template is="dom-if" if="[[!hasMissed]]">
-				<d2l-end-of-lesson-image completed=""></d2l-end-of-lesson-image>
+			<template is="dom-if" if="[[!hasMissed]]">
+			<d2l-end-of-lessons-progress href="{{_getRootHref(entity)}}" token="[[token]]">
```
ianlaidlaw: nit: this should be 1 tab to the right because it's nested below the `<template>`

components/d2l-end-of-lesson-progress.js
```diff
@@ -0,0 +1,41 @@
...
+						value="[[completionCount.completed]]"
+						max="[[completionCount.total]]"
+						style="width:12%">
```
ianlaidlaw: not a big deal, but should we use css here instead of inline styles? might be easier to visualize all the styling being done if it's not inside a the template here

components/d2l-end-of-lesson-progress.js
```diff
@@ -0,0 +1,40 @@
...
+				notify: true,
+				observer: '_scrollToTop'
+			},
+			token: {
+				type: String
+			}
+		};
+	}
+}
+customElements.define(D2LEndOfLessonsProgress.is, D2LEndOfLessonsProgress);
```
ianlaidlaw: These here should also drop the `s` from Lessons for consistency

components/d2l-end-of-lesson-progress.js
pbrown-d2l: We could for sure. That is just how the meter demo page showed it.
https://github.com/BrightspaceUI/core/blob/master/components/meter/demo/meter.html

pbrown-d2l: We should definitely do the best practice for sure. I will make this change as soon as my build works again

PR #181, opened 2020-02-13
DE37821 calcs suck

ChrisKraljevicD2L:
Turns out fixing the CSS for the sidenav involves changing stuff over here

### DIFF
--- a/d2l-sequence-navigator/d2l-sequence-navigator.js
+++ b/d2l-sequence-navigator/d2l-sequence-navigator.js
@@ -43,7 +43,7 @@ PolymerElement
 		}
 
 		.module-content {
-			height: calc( 100% - 259px );
+			height: calc( 100% - 203px );
 			border-top: 1px solid var(--d2l-color-mica);
 		}
 


d2l-sequence-navigator/d2l-sequence-navigator.js
```diff
@@ -43,7 +43,7 @@ PolymerElement
 		}
 
 		.module-content {
-			height: calc( 100% - 259px );
+			height: calc( 100% - 203px );
```
ianlaidlaw: why 203? does this number have any real significance?

ChrisKraljevicD2L: I removed the 56px `top` from the sidebar, so this is 259 - 56.

ianlaidlaw: sucks that this component has to be so aware of the parent using it.. I know that the launcher FRA uses this component, will shrinking this be an issue?

ChrisKraljevicD2L: I will test this out in sequence launcher. I have a sinking feeling that it hasn't been updated in a very long time.

PR #194, opened 2020-03-17
Prevent children from rendering when accordion starts closed

ChrisKraljevicD2L:
We don't need extra network calls being made from child components of modules before the module's open. After it's open those children can stay in the DOM, no need to re-render them each time the accordion opens/closes

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
@@ -176,7 +176,7 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		</style>
 
 		<d2l-labs-accordion-collapse no-icons="" flex="">
-			<div slot="header" id="header-container" class$="[[_getIsSelected(currentActivity, focusWithin)]] [[isEmpty(subEntities)]] [[_getHideDescriptionClass(_hideModuleDescription, isSidebar)]]" is-sidebar$="[[isSidebar]]">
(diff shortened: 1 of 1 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -273,6 +273,10 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 			_hideModuleDescription: {
 				type: Boolean,
 				computed: '_getHideModuleDescription(entity)'
+			},
+			_allowChildrenRendering: {
+				type: Boolean,
+				value: false
```
ChrisKraljevicD2L: Right now all modules are starting closed by default, eventually there'll be logic in here to determine whether or not a module contains the "current activity" and should start open.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -238,7 +238,7 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 			},
 			subEntities: {
 				type: Array,
-				computed: 'getSubEntities(entity)'
+				computed: 'getSubEntities(entity, _allowChildrenRendering)'
```
ianlaidlaw: I think this is fine to merge for now cause we're working so incrementally, but the approach I have in my head might run into a problem by having subEntities depend on `_allowChildrenRendering`.

At this point in the flow, I can get the current activity entity, and then I'll need to determine if it's a child of this `d2l-sequence-launcher-module`. I'll need the subEntities so that I can check if my current activity href is equal to any of the subEntities hrefs (or this href) or if the current activities "up" href that I get off the entity is equal to any of these (or this href).

ianlaidlaw: All that is to say we'll need to get subEntities so I can use them, but still prevent them from being rendered and making requests

PR #196, opened 2020-03-19
Don't render a module's children unless it is visible. (Lazy Loading)

ianlaidlaw:
sequence lawn chair pr: https://github.com/Brightspace/sequence-launcher-fra/pull/103

This pr provides lazy loading at the Lesson / Module level. How it works is that it will wrap it's children in a `dom-if` whose condition is determined by 2 things:
1. If the last accessed module is inside it (checks 2 levels)
2. If the user clicks the header

Note for 1. is that the accordion for the module will start visibly closed even though I can see it is correctly evaluating to true. So those elements are being fetched/rendered but you still need to click to see them. (For now, implemented in a future pr)

This pr does not have any skeletons or anything, so those topics will pop-in when expanded much like the lazy loading for the unit that was done a while ago.

I also did some small minor cleanup.

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
@@ -294,6 +294,10 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		};
 	}
 
+	static get observers() {
+		return ['_getShowModuleChildren(_moduleStartOpen, _moduleWasExpanded)'];
+	}
+
 	_accordionCollapseClass(focusWithin) {
 		return this._focusWithinClass(focusWithin);
 	}
@@ -423,8 +427,8 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		this._moduleWasExpanded = true;
 	}
 
-	_getShowModuleChildren() {
-		return this._moduleStartOpen || this._moduleWasExpanded;
+	_getShowModuleChildren(_moduleStartOpen, _moduleWasExpanded) {
+		return _moduleStartOpen || _moduleWasExpanded;
 	}
 
 	childIsActiveEvent() {
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
@@ -175,6 +175,7 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 
 		</style>
 
+		<siren-entity href="[[href]]" token="[[token]]" entity="{{_currentActivityEntity}}"></siren-entity>
 		<d2l-labs-accordion-collapse no-icons="" flex="">
 			<div slot="header" id="header-container" class$="[[_getIsSelected(currentActivity, focusWithin)]] [[isEmpty(subEntities)]] [[_getHideDescriptionClass(_hideModuleDescription, isSidebar)]]" is-sidebar$="[[isSidebar]]">
 				<div class="bkgd"></div>
@@ -238,7 +239,7 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 			},
 			subEntities: {
 				type: Array,
-				computed: 'getSubEntities(entity, _allowChildrenRendering)'
+				computed: 'getSubEntities(entity)'
 			},
 			hasChildren: {
 				type: Boolean,
@@ -276,7 +277,10 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 			},
 			_allowChildrenRendering: {
 				type: Boolean,
-				value: false
+				value: '_getAllowChildrenRendering(entity, subEntities, _currentActivityEntity)'
+			},
+			_currentActivityEntity: {
+				type: Object
 			}
 		};
 	}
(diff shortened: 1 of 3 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -268,19 +271,33 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
 				computed: '_getHideModuleDescription(entity)'
 			},
-			_allowChildrenRendering: {
+			lastViewedContentObject: {
+				type: String
+			},
+			_lastViewedContentObjectEntity: {
+				type: Object
+			},
+			_moduleStartOpen: {
```
ianlaidlaw: I don't love all the names I came up with for this. I'm open to suggestions

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -268,19 +271,33 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+			},
+			_moduleWasExpanded: {
 				type: Boolean,
 				value: false
 			}
 		};
 	}
 
+	static get observers() {
+		return ['_getShowModuleChildren(_moduleStartOpen, _moduleWasExpanded)'];
```
ianlaidlaw: I made this controlled by 2 booleans because Polymer didn't like that I had 1 set by a computed property and then tried to manually override it somewhere else. I think this is better anyways. This observer is so the dom-if gets picked correctly up on change of either of these

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -365,6 +382,17 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		return this._getTrueClass(focusWithin, selected);
 	}
 
+	_getModuleStartOpen(entity, subEntities, _lastViewedContentObjectEntity) {
+		if (!entity || !_lastViewedContentObjectEntity) {
+			return false;
+		}
+
+		const lastViewedParentHref = _lastViewedContentObjectEntity.getLinkByRel('up').href;
+		const thisHref = this.entity && this.entity.getLinkByRel('self').href;
```
ianlaidlaw: I think this component can just do `this.href`. I'll check

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -179,18 +163,7 @@ PolymerElement
...
 
 	isLast(entities, index) {
-		if (entities.length <= index + 1) {
-			return true;
-		}
-		else {
-			return false;
-		}
-	}
-	isSidebar() {
```
ianlaidlaw: I think I'll bring this stuff back for now. Eventually it won't need it, but I don't want to muddy this pr too much or cause unintended side-effects.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ChrisKraljevicD2L: SGTM :+1: (sounds good to me)

ianlaidlaw: (Sarah, get this merged.)

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -365,6 +382,17 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+
+		return lastViewedParentHref === thisHref || subEntities.some((s) => s.href === lastViewedParentHref);
```
ianlaidlaw: LMK if these needs some comments, or split it out into variables with names to describe scenarios.

The left hand is basically "the topic is directly under the module"
The right is "one of this modules children (which is an inner module) has the current activity as a child"

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
sarahboyd: lol took me a second to get this, was trying to figure out what I hadn't merged 😆

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ChrisKraljevicD2L: Is there a reason why `this.entity` is being used here instead of the `entity` parameter?
Also, can't the `href` property be used instead?

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ChrisKraljevicD2L: Is it possible that `_lastViewedContentObjectEntity`'s `self` link could equal `thisHref`, or was I incorrect in assuming that?

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: Nope, no reason. I just copied this line from elsewhere early on, but yeah I can just use `href` I think. Gonna change that

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: Ah, yeah it can be. In that case we should expand the module. I'll add that check in as well

PR #198, opened 2020-03-20
Open module containing last viewer content

ChrisKraljevicD2L:
I ain't gonna be able to test this for a while since I gotta rebuild my LMS, but this is the gist of the changes

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
@@ -495,10 +495,9 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		return _hideModuleDescription && !isSidebar ? 'hide-description' : '';
 	}
 
-	_openModule(moduleStartOpen) {
-		if (moduleStartOpen) {
+	_openModule(_moduleStartOpen) {
+		if (_moduleStartOpen) {
 			this.shadowRoot.querySelector('d2l-labs-accordion-collapse').setAttribute('opened', '');
-			this._moduleWasExpanded = true;
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -491,5 +494,12 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 	_getHideDescriptionClass(_hideModuleDescription, isSidebar) {
 		return _hideModuleDescription && !isSidebar ? 'hide-description' : '';
 	}
+
+	_openModule(moduleStartOpen) {
```
ianlaidlaw: nit: for consistency, this should be kept as `_moduleStartOpen` I think

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -491,5 +494,12 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+		if (moduleStartOpen) {
+			this.shadowRoot.querySelector('d2l-labs-accordion-collapse').setAttribute('opened', '');
+			this._moduleWasExpanded = true;
```
ianlaidlaw: I might be wrong, but I don't think we need to set this but I suppose it doesn't hurt

ChrisKraljevicD2L: I _think_ you're right, but it is used in `_getShowModuleChildren` so I'd rather set it to true to be extra safe

ChrisKraljevicD2L: Oh it's `OR`'d. I stand corrected

PR #207, opened 2020-04-06
Redesign the Launcher: Part 1

ianlaidlaw:
This pr is part 1 of the redesign of LX. I want to get some eyes on this sooner rather than later.

Things changed:
- The old design
- The hover states
- Gets rid of the "current" selected item that was used in the sidebar. This is easy to add back in, but it appears there is no design for this yet. The process was overcomplicated, all we should need to do in the future is check for the current href and compare.
- Redesigns submodules but does not add in collapsible state yet

image

Screenshot_20200406-164430_Chrome


ianlaidlaw:
**KNOWN ISSUES:**
- [ ] Skeletons on firefox
- [ ] Launch module button not working if not opened in new tab
- [ ] submodule has no collapse state
- [ ] the collapse/expand arrows are positioned differently

These may or may not be fixed in this pr, depending on what it is.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -202,10 +157,21 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 								[[localize('sequenceNavigator.optional')]]
 							</span>
 						</template>
-					</span>
+						<template is="dom-if" if="true">
```
ChrisKraljevicD2L, line 160: I'm guessing the `true` in this `dom-if` is getting replaced?

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -360,6 +330,36 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
 		return link && link.hasClass('sequenced-activity');
 	}
 
+	_getStartingCollapseIconName(entity, subEntities, _lastViewedContentObjectEntity) {
+		return this._getModuleStartOpen(entity, subEntities, _lastViewedContentObjectEntity)
+			? 'tier1:arrow-collapse-small'
+			: 'tier1:arrow-expand-small';
+	}
+
+	_activitySiblingIsActivity(subEntities, index) {
```
ianlaidlaw, line 339: This fn name is a little awkward. but basically I need to know if it's next sibling is an activity, cause I may or may not show a grey underline.

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -25,41 +24,9 @@ class D2LActivityLink extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Completi
...
-			:host(.d2l-asv-focus-within),
-			:host(:focus),
-			:host(:hover) {
-				--d2l-activity-link-background-color: var(--d2l-asv-primary-color);
-				--d2l-activity-link-subtext-color: var(--d2l-asv-text-color);
-				--d2l-activity-link-border-color: rgba(0, 0, 0, 0.42);
-				--d2l-activity-link-text-color: var(--d2l-asv-text-color);
-				--d2l-activity-link-opacity: 0.26;
-				--d2l-activity-link-backdrop-opacity: 1;
+				padding-left: 24px;
```
ianlaidlaw, line 28: ideally we'll get these all into vars or something.. once we nail down the approach

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -194,29 +124,39 @@ class D2LActivityLink extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Completi
...
+			<template is="dom-if" if="[[showLoadingSkeleton]]">
+				<div id="skeleton" class="skeleton"></div>
+			</template>
+			<template is="dom-if" if="[[!showLoadingSkeleton]]">
+				<div id="content-container" on-click="_contentObjectClick">
+					<template is="dom-if" if="[[hasIcon]]">
+						<d2l-icon icon="[[_getIconSetKey(entity)]]"></d2l-icon>
+					</template>
+					<div class="d2l-activity-link-title">
+						<a on-click="setCurrent" class$="[[completionRequirementClass]]" href="javascript:void(0)">
```
ianlaidlaw, line 151: I think I can get rid of all this "setCurrent" fucntions and events and just simply make this an a tag with an href

d2l-sequence-navigator/d2l-inner-module.js
```diff
@@ -183,13 +108,14 @@ class D2LInnerModule extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Completio
...
-			<div id="module-header" class$="[[_getIsSelected(currentActivity, focusWithin)]] [[_getHideDescriptionClass(_hideDescription)]]" on-click="_onHeaderClicked">
-				<div class="bkgd"></div>
-				<div class="bkgd-backdrop"></div>
-				<div class="border"></div>
-				<a on-click="_onHeaderClicked" href="javascript:void(0)">
-					<span class="module-title">[[entity.properties.title]]</span>
-				</a>
+			<div id="module-header" class$="[[[[_getHideDescriptionClass(_hideDescription)]]" on-click="_onHeaderClicked">
+				<div on-click="_onHeaderClicked" style="width: 100%;">
+					<div style="display: inline-flex;">
```
ianlaidlaw, line 113: move these inline styles up to selectors

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -360,6 +330,36 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+		if (index >= subEntities.length) {
+			return false;
+		}
+
+		const nextSibling = subEntities[index + 1];
+
+		return !this._isActivity(nextSibling);
```
ChrisKraljevicD2L, line 346: Couple things here.
Are we ok with indexing on `subEntities.length`, when `index` equals `subEntites.length - 1`?
And, it's a bit confusing to name the function `_activitySiblingIsActivity` and return `!this._isActivity(nextSibling)`. i.e. if `nextSibling` is an activity (`_isActivity` returns `true`), this function would return `false`.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -360,6 +330,36 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+
+	// TODO: this function needs work
+	_getLaunchModuleHref(entity) {
+		if (!entity) {
+			return '';
+		}
+
+		if (entity.getLinkByRel('alternate')) {
+			return entity.getLinkByRel('alternate').href;
+		}
```
ChrisKraljevicD2L, line 357: Why the `alternate` href? iirc it's because that href has the LMS as its host but that's not immediately obvious to someone reading the code. Maybe a comment here saying "the sequences api sucks"?

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -2,90 +2,45 @@ import '../d2l-sequence-navigator/d2l-inner-module.js';
...
 		<style>
 			:host {
 				display: block;
 				@apply --d2l-body-compact-text;
-				width: 100%;
 				--d2l-outer-module-text-color: var(--d2l-asv-text-color);
 				--d2l-outer-module-background-color: transparent;
 				--d2l-activity-link-padding: 10px 24px;
-				margin-top: -1px;
+				border: 1px solid var(--d2l-color-mica);
```
ianlaidlaw, line 27: I think I will need to take this off host and put it on an actual element so it won't show when the skeleton is showing.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -2,90 +2,45 @@ import '../d2l-sequence-navigator/d2l-inner-module.js';
...
+				border-radius: 8px;
```
ianlaidlaw, line 28: 6px

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -78,10 +74,24 @@ PolymerElement
...
+									token="[[token]]"
+									current-activity="{{href}}"
+									is-sidebar="[[isSidebar()]]"
+									last-module="[[isLast(subEntities, index)]]"
+									last-viewed-content-object="[[lastViewedContentObject]]"
+								>
+								</d2l-sequence-launcher-module>
 							</template>
 							<template is="dom-if" if="[[_isActivity(childLink)]]">
-								<d2l-activity-link href="[[childLink.href]]" token="[[token]]" current-activity="{{href}}" before-module$="[[isBeforeModule(subEntities, index)]]"></d2l-activity-link>
```
ianlaidlaw, line 84: idk what this `before-module` attribute did... I could not find anything

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -171,13 +181,14 @@ PolymerElement
...
-			}
+	_activitySiblingIsActivity(subEntities, index) {
+		if (index >= subEntities.length) {
+			return false;
 		}
-		return false;
+
+		const nextSibling = subEntities[index + 1];
+
+		return !this._isActivity(nextSibling);
```
ChrisKraljevicD2L, line 191: Same comment here as in sequence launcher module

d2l-sequence-navigator/d2l-inner-module.js
```diff
@@ -256,26 +183,33 @@ class D2LInnerModule extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Completio
...
 
+	_activitySiblingIsActivity(subEntities, index) {
+		if (index >= subEntities.length) {
+			return false;
+		}
+
+		const nextSibling = subEntities[index + 1];
+
+		return !this._isActivity(nextSibling);
+	}
```
ChrisKraljevicD2L, line 194: Anotha one

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: whoops, good catch. originally i was trying to render the different icon names using dom-if but that wasn't working, this looks like leftover. i can delet this dom-if

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: for point 1. are you suggesting that I need to improve the initial if check `index >= subentities.length` to be `index + 1` ? I think you might be right.

2. Yeah this is all very poorly named. I'll try and come up with an improvement. I think I had reversed the logic at some point but kept the name when I was testing so that might explain that. Either way, it's bad - will improve.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: Yeah... pretty much. `alternate` seemed to be the only one that was working. You can see my comment below about the `self` link giving me a 401, and it also just doesn't look right. I _should_ have the link to that item here somewhere, maybe I just need to get it from a better place.

But also `alternate` is only working right now if I open in a new tab, otherwise it keeps you on the launcher route. It's weird.

d2l-sequence-navigator/d2l-inner-module.js
ianlaidlaw: I hate rewriting functions like this. It's almost like I should use a mixin but I also really do not like those either... ugh. Maybe I could rewrite the logic into a helper function, but still have the base function on the observables? What do you think?
```
_wrapperFunction(subEntities, index) {
  return myHelperFunction(subEntities, index);
}
```

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: Either way, this function still needs work I'll have to address in this pr, or in the future.

d2l-sequence-navigator/d2l-inner-module.js
```diff
@@ -221,7 +148,7 @@ class D2LInnerModule extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Completio
 			},
 			_hideDescription: {
 				type: Boolean,
-				computed: '_getHideDesciption(entity)'
```
ianlaidlaw, line 224: when i read `desciption` for some reason all i see is `despacito`

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
ianlaidlaw: Actually I think it might be safe. If `index` is `6` and `length` is `6`, then `index + 1` would give us an error... But the first if check will catch that because of the `>=`
But maybe there's a way to rewrite this to be more clear

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -202,10 +157,19 @@ class D2LSequenceLauncherModule extends ASVFocusWithinMixin(PolymerASVLaunchMixi
...
+						<d2l-icon id="expand-icon" icon="[[_iconName]]"></d2l-icon>
+					</div>
 				</div>
 				<div id ="startDate">[[startDate]]</div>
 			</div>
+
+			<div id="launch-module-container">
+				<a href="[[_launchModuleHref]]">
+					<!--todo: get lang term and add a11y stuff-->
+					<d2l-button-subtle text="Launch Module" icon="tier1:move-to"></d2l-button-subtle>
```
sarahboyd, line 169: should we be adding a lang term for this before merging?

ianlaidlaw: yes, we will need to get those. do you know what the process looks like for getting that? I believe this is a new term

sarahboyd: you just need to add the new term to the en.json file and the automation will take it from there (we'll serge will send it off for translations and then give us those fun PRs)

d2l-sequence-navigator/d2l-inner-module.js
ChrisKraljevicD2L: For just one function I don't think a mixin is required. Maybe just export the function in its own js file and import it where ya need it. But either route, mixin or import, yeah a wrapper function would be good for readability.

d2l-sequence-navigator/d2l-activity-link.js
ianlaidlaw: not gonna worry about this right now

d2l-sequence-navigator/d2l-inner-module.js
ianlaidlaw: in the interest of moving faster, i'm not going to worry about this right now

PR #213, opened 2020-04-09
Fix submodule design, and other stuff

ianlaidlaw:
- Fixed up some stuff to get the implementation closer to design. Fixed a lot of things related to padding, spacing, etc. 
- Sub modules are now properly designed.
- Made things way less fragile by moving to things over to flexbox where possibel.

image

image

image


ianlaidlaw:
I accidentally clicked the close button I think? This should be open.

lang/en.json
```diff
@@ -29,7 +29,7 @@
 	"sequenceNavigator.currentModule": "Unit {current} of {total}",
 	"sequenceNavigator.due": "Due {dueDate}",
 	"sequenceNavigator.exempt":"Exempt",
-	"sequenceNavigator.launchModule": "Launch Module",
+	"sequenceNavigator.launchModule": "Launch Lesson",
```
ianlaidlaw, line 32: This copy was updated

demo/d2l-sequence-navigator/data/m2-activity3.json
```diff
@@ -1,6 +1,6 @@
 {
 	"properties": {
-		"title": "ThisIs1ReallyLongWordddddddddddddddddddddddddddddddddddddddddddddddddddd."
```
ChrisKraljevicD2L, line 3: Sorry to nitpick, but I think this is still a valid "test case", and we could make a new activity with a title like the new one instead of replacing this one.

pbrown-d2l: Not a huge issue, but yeah ideally adding a new case is better

ianlaidlaw: done.
image

I've checked the completion status with both types of names, it's in the right spot in both cases

PR #219, opened 2020-04-16
DE38307 - Using the langterm from the other half of an incomplete fea…

pbrown-d2l:
…ture


pbrown-d2l:
Putting this here while I test it locally. Requires https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/10737/overview

PR #221, opened 2020-04-22
Break words onto new lines

ianlaidlaw:
Things changed and `word-wrap` no longer did what it was supposed to do, so `word-break` was needed.

`word-break` breaks words against the width of the parent container, and since this stuff converted to using flexbox, this was more appropriate.

https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

image


ChrisKraljevicD2L:
Though, the tests in here are _extremely_ flakey

sarahboyd:
> Though, the tests in here are _extremely_ flakey

I've seen a lot of tests just timing out lately :( maybe something is up with sauce?

ChrisKraljevicD2L:
@ianlaidlaw my other PR's merged now, 🚢

PR #224, opened 2020-04-30
DE38773 completion entities don't count for unit early load

ChrisKraljevicD2L:
Previously, for some reason, if there were no children for a given unit its only subentity would be a completion entity. Now that's accounted for when checking whether or not a unit should fire off an early loading event when it has no children.

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -251,7 +251,11 @@ PolymerElement
 	}
 
 	_checkForEarlyLoadEvent(entity, subEntities) {
-		if (entity && subEntities && subEntities.length <= 0) {
+		if (entity &&
+			subEntities && (
+			subEntities.length <= 0 ||
+			subEntities[0].rel.includes('item'))
+		) {
 			this.dispatchEvent(new CustomEvent('d2l-content-entity-loaded', {detail: { href: this.href}}));
 		}
 	}
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -253,7 +253,7 @@ PolymerElement
 	_checkForEarlyLoadEvent(entity, subEntities) {
 		if (entity &&
 			subEntities && (
-			subEntities.length <= 0 ||
+			subEntities.length <== 0 ||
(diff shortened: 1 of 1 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -251,7 +251,11 @@ PolymerElement
 	}
 
 	_checkForEarlyLoadEvent(entity, subEntities) {
-		if (entity && subEntities && subEntities.length <= 0) {
+		if (entity &&
+			subEntities && (
+			subEntities.length <= 0 ||
+			subEntities[0].rel.includes('item'))
```
apalaniuk, line 257: Peanut gallery here: not knowing the exact shape of the (unknown, to me) subentities on this (unknown, to me) entity, it seems possible that other entities which are not collection items might be added at some point in the future, and these might come before a collection item. Is it possible that you want to `Array.prototype.find` to see if there are any collection items?

This might not be true, because I don't know what an early load event is in this context, just seems like it might not be resilient to future changes.

ChrisKraljevicD2L: You're right, this is a pretty fragile implementation. I'll use find :+1:

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -251,7 +251,11 @@ PolymerElement
 	}
 
 	_checkForEarlyLoadEvent(entity, subEntities) {
-		if (entity && subEntities && subEntities.length <= 0) {
+		if (entity &&
+			subEntities && (
+			subEntities.length <= 0 ||
```
ianlaidlaw, line 256: small nit: in this example I don't see the type conversion getting screwed up cause we'll know what `.length` is, but I would suggest we might as well use `<==` here

ChrisKraljevicD2L: I should have know there's a super special less than operator for js :P I'll get that fixed

ianlaidlaw: Yeah, it's definitely trivial in this example but just doesn't attempt to perform type conversion.. eg:
```
0 == '0'; // true
0 === '0'; // false
```
I imagine it's completely negligible but I would think that triple equals has faster performance as well since it doesn't check those conversions. 🤷

PR #226, opened 2020-05-07
DE34552 end of lesson activities not focusing via keyboard

imcdonald-d2l:
While the `focusWithin` property was being updated properly in `ASVFocusWithinMixin` it was not setting the class on the `d2l-eol-activity-link` component. This was fixed by adding the class property to the component and reflecting it as an attribute to add the class on focus.

### DIFF
--- a/d2l-sequence-navigator/d2l-eol-activity-link.js
+++ b/d2l-sequence-navigator/d2l-eol-activity-link.js
@@ -156,6 +156,11 @@ class D2LEolActivityLink extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Compl
 	}
 	static get properties() {
 		return {
+			class: {
+				type: String,
+				reflectToAttribute: true,
+				computed:'_focusWithinClass(focusWithin)'
+			},
 			hasIcon: {
 				type: Boolean,
 				computed: '_hasIcon(entity)'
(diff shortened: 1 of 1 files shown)


d2l-sequence-navigator/d2l-eol-activity-link.js
```diff
@@ -156,6 +156,11 @@ class D2LEolActivityLink extends ASVFocusWithinMixin(PolymerASVLaunchMixin(Compl
 	}
 	static get properties() {
 		return {
+			class: {
+				type: String,
+				reflectToAttribute: true,
+				computed:'_focusWithinClass(focusWithin)'
+			},
```
ChrisKraljevicD2L, line 163: I'm slightly concerned with this approach. Mostly because consumers of `d2l-eol-activity-link` could safely assume they can set the `class` attribute and not have it overwritten.

A better approach may be to use the logic inside `_focusWithinClass` to add/remove a class from `classList` and turn it into an observer on `focusWithin`. I'm gonna do some searching to see if this is possible.

ChrisKraljevicD2L: Yup, `this.classList` works just fine along with `.add`, `.remove` and `.toggle`.
To set up your observer, you'll need a static getter like in [`d2l-sequences-content-eol-main.js`](https://github.com/BrightspaceHypermediaComponents/sequences/blob/master/components/d2l-sequences-content-eol-main.js#L151) on line 151.
Then make a function that will `.add` or `.remove` the `d2l-asv-focus-within` class based on the value of `focusWithin`.
Finally add that function to your observers and you should be good :+1:
Feel free to reach out over slack or for a quick zoom call to help with any of this.

imcdonald-d2l: Thanks @ChrisKraljevicD2L! I'll look into that and reach out to you if I have further questions.

PR #228, opened 2020-05-08
imcdonald/de38770 dynamic width of eol activity links to account for mobile devices

imcdonald-d2l:
Used media query to dynamically set width of content so it is easily visible on a mobile device.
Demo page for testing added to `demo/d2l-sequence-navigator/d2l-sequences-content-eol-main.html`. You can open dev tools and set device to try various display widths.
[Original Rally Ticket](https://rally1.rallydev.com/#/289692574792d/detail/defect/387156470440)

Before:
before

After:
after

### DIFF
--- a/components/d2l-sequences-content-eol-main.js
+++ b/components/d2l-sequences-content-eol-main.js
@@ -43,7 +43,9 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 				margin: auto!important;
 			}
 			@media (max-width:550px) {
-				d2l-missed-activity { width: 100%; }
+				d2l-missed-activity {
+					width: 100%;
+				}
 			}
 		</style>
 		<div class="content-eol-main-container">
--- a/components/d2l-sequences-content-eol-main.js
+++ b/components/d2l-sequences-content-eol-main.js
@@ -37,10 +37,13 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 			.missed-count-text {
 				color: var(--d2l-color-celestine);
 			}
-			d2l-missed-activity{
+			d2l-missed-activity {
 				text-align: left;
 				width: 50%;
 				margin: auto!important;
+			}
+			@media (max-width:550px) {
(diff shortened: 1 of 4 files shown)


components/d2l-sequences-content-eol-main.js
```diff
@@ -37,11 +37,14 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 			.missed-count-text {
 				color: var(--d2l-color-celestine);
 			}
-			d2l-missed-activity{
+			d2l-missed-activity {
 				text-align: left;
 				width: 50%;
 				margin: auto!important;
 			}
+			@media (max-width:550px) {
```
ChrisKraljevicD2L, line 45: nit: I'm a bit fuzzy on this, but I'm not sure if this is our mobile width breakpoint. Either [design.d2l](http://design.d2l) will have it listed somewhere in there, or you can reach out to Tara and ask :+1:

ianlaidlaw: yeah, the mobile breakpoint for us (and I think standard) is 929px. Here's a screenshot from one of our design docs:
image

components/d2l-sequences-content-eol-main.js
```diff
@@ -37,11 +37,14 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
...
+				d2l-missed-activity { width: 100%; }
```
ianlaidlaw, line 46: very small nit: generally we've been spacing these out even if it's just a single item.
```
d2l-missed-activity {
  width: 100%;
}
```

components/d2l-sequences-content-eol-main.js
imcdonald-d2l: So if we set `929px` as breakpoint we get grossness below if the user's browser window is smaller. Not sure if we're ok with this?

also, what does `nit` stand for 😅 

smaller-desktop

components/d2l-sequences-content-eol-main.js
imcdonald-d2l: Thanks for letting me know 👍

PR #232, opened 2020-05-19
imcdonald/US116703 Lesson header style changes and performance changes

imcdonald-d2l:
Styling `d2l-lesson-header` so it reflects updates in sequence navigator designs outlined [here](https://xd.adobe.com/view/e16d4c3f-77f8-4632-6a01-a98beb7c5072-975c/screen/2fc07bba-3928-4455-a3c2-589589994d8d/Specs-Navigation-Button-States/).

Also adjusting `moduleIndex` and `siblingModules` so they fetch information from the new location with Al's performance changes.

### DIFF
--- a/d2l-sequence-navigator/d2l-lesson-header.js
+++ b/d2l-sequence-navigator/d2l-lesson-header.js
@@ -25,6 +25,8 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 			--d2l-lesson-header-text-color: var(--d2l-asv-text-color);
 			--d2l-lesson-header-background-color: transparent;
 			--d2l-meter-size: 48px;
+			width: calc(100% - 14px);
+			height: calc(100% - 14px);
 			background-color: var(--d2l-lesson-header-background-color);
 			color: var(--d2l-lesson-header-text-color);
 			margin: 10px var(--d2l-sequence-nav-padding) 10px var(--d2l-sequence-nav-padding);
@@ -32,7 +34,6 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 			display: block;
 			position: relative;
 			z-index: 0;
-			width: 100%;
 		}
 
 		:host(.d2l-asv-current) div.border {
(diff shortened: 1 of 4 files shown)


d2l-sequence-navigator/d2l-lesson-header.js
```diff
@@ -282,7 +275,10 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 			},
 			_completionProgress: {
 				type: String,
-				computed: '_getCompletionProgress(entity.properties, _self)'
+				computed: '_getCompletionProgress(entity.properties, _self, _moduleIndex, _siblingModules)'
```
ChrisKraljevicD2L, line 278: I see that you've changed the call for `_getCompletionProgress` here, but the function signature hasn't changed. Did you want to prevent `_completionProgress` from being set until the module progress info had been fetched?

imcdonald-d2l: Yeah, basically I found that `_getCompletionProgress()` was not being re-computed when `_moduleIndex` and `_siblingModules` were being set so the proper information was not being shown in the UI. I fixed that by adding it as a dependency. My change wasn't to wait until module progress info has been fetched but would that be preffered?

ChrisKraljevicD2L: Ah, I see what you were going for. Naw your approach is a-ok, no changes needed :+1:

test/d2l-sequence-navigator/d2l-lesson-header.html
```diff
@@ -193,20 +193,6 @@
...
-		});
-		it('should use the referenced lang term', async() => {
-			const unitInfo = element.shadowRoot.querySelector('.unit-info :last-child');
-			expect(unitInfo.innerText)
-				.to
-				.equal('Unit 2 of 4');
-		});
-
-	});
-
```
imcdonald-d2l, line 209: Above test was deleted due to difficulty in re-working a robust functioning test for this particular case. Unit information is being fetched from a secondary href as part of AL's changes and the asynchronous information makes this test difficult. Discussions were had with @ChrisKraljevicD2L and we thought it would be easier to remove this test than finding a robust/working solution.

ChrisKraljevicD2L:
This PR is good to go, but one small suggestion would be to add the same data that was added for tests to the demo as well

imcdonald-d2l:
> Does this still work with the old API?

Good question @alamarre, currently the way I have it setup is so that it would only work with your new changes. Do we want it backwards compatible? Shouldn't be too much work to do that

alamarre:
> > Does this still work with the old API?
> 
> Good question @alamarre, currently the way I have it setup is so that it would only work with your new changes. Do we want it backwards compatible? Shouldn't be too much work to do that

I think we should if it's not too hard so that if for whatever reason we needed to revert the API we could just flip the feature flag and not need to hotfix the UI.

PR #233, opened 2020-05-21
DE39021 - Adding iframe permissions

pbrown-d2l:


components/d2l-sequences-content-router.js
```diff
@@ -114,8 +114,8 @@ function getLinkEntityType(linkActivity) {
 	if (link && isLorPdf) {
 		return D2LSequencesContentFilePdf.is;
 	} else if (
-		(link && link.href.startsWith(window.location.protocol)) ||
-		(embedLink && embedLink.href.startsWith(window.location.protocol))
+		(link && link.href.startsWith('https')) ||
+		(embedLink && embedLink.href.startsWith('https'))
 	) {
```
pbrown-d2l, line 119: This is not quite correct. I will likely be reverting this shortly. This comes up during testing locally since your local build is http and it will stop you from seeing https links

PR #234, opened 2020-05-29
Align sidenav components to design spec

imcdonald-d2l:
Align left navigation to design spec. This PR tweaks the styles of various components to align with styles outlined on the link below. Main changes are the hover and focus states of component. Accordion changes were made so a version bump was required in `package.json`.

Links:
- [Rally Ticket](https://rally1.rallydev.com/#/289692574792d/detail/userstory/393476310436?fdp=true)
- [Design Spec](https://xd.adobe.com/view/e16d4c3f-77f8-4632-6a01-a98beb7c5072-975c/screen/2fc07bba-3928-4455-a3c2-589589994d8d/Specs-Navigation-Button-States?x_product=cc-slack%2F1.5.0)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   "main": "d2l-sequences.js",
   "dependencies": {
     "@brightspace-ui/core": "^1",
-    "@brightspace-ui-labs/accordion": "^2.0.2",
+    "@brightspace-ui-labs/accordion": "^2.3.0",
     "@d2l/audio": "Brightspace/d2l-audio#semver:^3",
     "@d2l/video": "Brightspace/d2l-video#semver:^1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
   "main": "d2l-sequences.js",
   "dependencies": {
     "@brightspace-ui/core": "^1",
-    "@brightspace-ui-labs/accordion": "^2.0.2",
+    "@brightspace-ui-labs/accordion": "^2.3.0",
```
imcdonald-d2l, line 67: Is this all that needs to be done to get the new version in properly? No extra work needed in the accordion component?

sarahboyd: I was gonna say you need to update the package-lock, but sequences doesn't have one so yes that should be it :)

PR #240, opened 2020-06-17
Adjust LX header so it says "lesson" instead of "unit" when useLessonAsRoot configured

imcdonald-d2l:
Adjust `_completionProgress` conditionals so that the text is properly set. The "lesson" lang term is passed in `moduleProperties.completionProgressLangTerm` on the back-end only when `useLessonAsRoot` is configured.

### DIFF
--- a/d2l-sequence-navigator/d2l-lesson-header.js
+++ b/d2l-sequence-navigator/d2l-lesson-header.js
@@ -262,7 +262,7 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 			},
 			_completionProgress: {
 				type: String,
-				computed: '_getCompletionProgress(entity.properties, _moduleIndex, _siblingModules)'
+				computed: '_getCompletionProgress(entity.properties, _moduleProgress.properties, _moduleIndex, _siblingModules)'
 			},
 			_moduleProgress: {
 				type: Object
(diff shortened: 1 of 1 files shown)


d2l-sequence-navigator/d2l-lesson-header.js
```diff
@@ -309,10 +309,16 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 		return;
 	}
 
-	_getCompletionProgress(properties) {
-		return properties && properties.completionProgressLangTerm
-		|| this._moduleIndex && this._siblingModules && this.localize('sequenceNavigator.currentModule', 'current', this._moduleIndex, 'total', this._siblingModules)
-		|| '';
+	_getCompletionProgress(entityProperties, moduleProperties, _moduleIndex, _siblingModules) {
+		if (entityProperties && entityProperties.completionProgressLangTerm) {
```
ianlaidlaw, line 313: not a big deal but destructuring could go a long way for readability here.
Maybe this isn't better but I'm trying to play around with something:
```
const { completionProgressLangTerm: entitityLangTerm } = entityProperties || {};
const { completionProgressLangTerm: moduleLangTerm } = moduleProperties || {};

if (entityLangTerm) {
  return entityLangTerm;
} else if (moduleLangTerm) {
  return moduleLangTerm;
}
```

the more I write this, the less I like it. I'm trying to avoid those `if x, return x ... else continue` statements but i don't see a great way around that except for evaluating all options at the beginning. Maybe you're fine to keep what you have

imcdonald-d2l: Yeah, I don't know if there is a perfect clean solution here. The lang term evaluation is overly complex given its various location and no easy way around it.

PR #241, opened 2020-06-18
Fix import on error page

ianlaidlaw:
This looks like it never worked. Importing the class component allows you to access elements on it.

image

This is the expected view when there is an error. Previously it would continue to try and show the content you just had, or something. If this file thinks there's an error we should probably show the error page...

image


ianlaidlaw:
This isn't really an urgent fix or anything. But I want to get some eyes on it since it definitely doesn't work in it's current state.

PR #242, opened 2020-07-02
Fix accordion collapse issue

imcdonald-d2l:
- Refactored some property logic that was a bit messy. Ultimately `_moduleWasExpanded` wasn't being set properly but some investigation found that it isn't necessary as the `accordionState` holds the same information. Refactored to remove `_moduleWasExpanded` property and reference `accordionState` instead.
- Confirmed with Tara and double tap to close is the expected behavior when clicking the module header (unless you are navigating to the module header from within another activity in the module)
module double click

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
@@ -575,7 +575,7 @@ class D2LSequenceLauncherModule extends PolymerASVLaunchMixin(CompletionStatusMi
 	}
 
 	_onHeaderClicked() {
-		this._moduleWasExpanded = true;
+		this._moduleWasExpanded = !this._moduleWasExpanded;
 		if (!this._hideModuleDescription && this.isSidebar) {
 			this.currentActivity = this.entity.getLinkByRel('self').href;
 			this._contentObjectClick();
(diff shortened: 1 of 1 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -536,8 +537,6 @@ class D2LSequenceLauncherModule extends PolymerASVLaunchMixin(CompletionStatusMi
 		if (!entity || (!_lastViewedContentObjectEntity && !_currentActivityEntity)) {
 			return false;
 		}
-		// Set the starting icon depending on the collapse state
-		this._updateCollapseStateAndIconName();
```
imcdonald-d2l, line 540: This was moved to `ready()` function as it only needs to be run when the component is ready, not every time `_getModuleStartOpen()` is executed.

d2l-sequence-launcher-unit/d2l-sequence-launcher-module.js
```diff
@@ -575,15 +574,17 @@ class D2LSequenceLauncherModule extends PolymerASVLaunchMixin(CompletionStatusMi
...
 		if (!this._hideModuleDescription && this.isSidebar) {
 			this.currentActivity = this.entity.getLinkByRel('self').href;
 			this._contentObjectClick();
 		}
 	}
 
-	_getShowModuleChildren(_moduleStartOpen, _moduleWasExpanded) {
-		return _moduleStartOpen || _moduleWasExpanded;
+	_getShowModuleChildren(_moduleStartOpen, accordionState) {
+		if (_moduleStartOpen || accordionState === 'open') {
```
ianlaidlaw, line 584: small nit, this could be a one liner or a ternary if you wanted to be a little cleaner

imcdonald-d2l: Good catch 👍

PR #245, opened 2020-07-03
update d2l-video

imcdonald-d2l:
Update `d2l-video` version. Updated sequence version and BSI `d2l-video` update to follow merge so it works locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@ node_modules
 locales.json
 .vs
 analysis.json
-package-lock.json
 build
--- a/package.json
+++ b/package.json
@@ -63,10 +63,10 @@
   },
   "main": "d2l-sequences.js",
   "dependencies": {
-    "@brightspace-ui/core": "^1",
+    "@brightspace-ui/core": "^1.53.0",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -63,10 +63,10 @@
   },
   "main": "d2l-sequences.js",
   "dependencies": {
-    "@brightspace-ui/core": "^1",
+    "@brightspace-ui/core": "^1.53.0",
```
ianlaidlaw, line 66: same question here, any reason to move away from `"^1"` ?

imcdonald-d2l: I (improperly) assumed it needed the version updated. See comment on lessons PR for more context. Will update this as well

.gitignore
```diff
@@ -2,5 +2,4 @@ node_modules
 locales.json
 .vs
 analysis.json
-package-lock.json
```
ChrisKraljevicD2L, line 5: Is there a reason why `package-lock` is being reinstated here?

imcdonald-d2l: Travis build wouldn't pass without it for some reason. Not sure if there is a way around this 🤷‍♂️

PR #246, opened 2020-07-07
Enabling dependabot

ChrisKraljevicD2L:
This is technically part of the killing sequence viewer stuff, but I figured it should be its own PR

Also dunno if this should be done or not lol

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,8 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "daily"
+    allowed_updates:
+      - match:
+          update_type: "security"


.dependabot/config.yml
```diff
@@ -0,0 +1,8 @@
+version: 1
```
dbatiste, line 1: I very recently set this up for one of my projects.  Note [the documentation](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem) suggests `version: 2` is required.

This is what I used yesterday when I was using this:
```
# Basic dependabot.yml file with minimum configuration for npm

version: 2
updates:
  # Enable version updates for npm
  - package-ecosystem: "npm"
    # Look for `package.json` and `lock` files in the `root` directory
    directory: "/"
    # Check the npm registry for updates every day (weekdays)
    schedule:
      interval: "daily"
```
Is there something being used here that relies on version 1?

dbatiste: I think there are some subtle differences in the API, as well as the expected location for the file.  This is the one I had created: https://github.com/BrightspaceUI/stylelint-config/blob/master/.github/dependabot.yml

PR #247, opened 2020-07-09
Modify completion check behavior to get (faster) real time updates

imcdonald-d2l:
Previously completion check only occurred upon navigating to a new item in the nav bar (or through arrow buttons). This didn't allow us to update completion count or completion status in a more real-time manner. For example, when viewing HTML topics, they are considered complete just after viewing them. This means we can update completion count and status immediately after opening the link. However, some items like discussion, assignments, surveys etc. shouldn't be polled for up updates until navigating out of the item since we have no way of knowing when a user has completed the activity. The code changes below account for these scenarios and try to better update completion count in a more real-time manner.

Gif below shows immediate completion count updates on items that are marked as complete on view.
completion-count

### DIFF
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -178,7 +178,7 @@ class D2LSequenceLauncherUnit extends D2L.Polymer.Mixins.Sequences.CompletionTra
 
 	_onHrefChanged(href, previousHref) {
 		if (previousHref && previousHref !== href && this._previousEntity) {
-			this.finishPreviousEntityCompletion(this._previousEntity);
+			this.startPreviousEntityCompletion(this._previousEntity);
 		}
 	}
 
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -1,26 +1,17 @@
 import '@polymer/polymer/polymer-legacy.js';
-import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
 import './d2l-sequence-launcher-module.js';
 import '@brightspace-ui-labs/accordion/accordion.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import 'siren-entity/siren-entity.js';
-import '../localize-behavior.js';
 import '../mixins/d2l-sequences-completion-tracking-mixin.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
-import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
-import { PolymerElement } from '@polymer/polymer/polymer-element.js';
+
 /*
 @memberOf D2L.Polymer.Mixins;
 @mixes SirenEntityMixin
 */
 
-class D2LSequenceLauncherUnit extends mixinBehaviors([
-	D2L.PolymerBehaviors.Siren.EntityBehavior,
-	D2L.Polymer.Mixins.Sequences.CompletionTrackingMixin,
-	D2L.PolymerBehaviors.Sequences.LocalizeBehavior
-],
-PolymerElement
-) {
+class D2LSequenceLauncherUnit extends D2L.Polymer.Mixins.Sequences.CompletionTrackingMixin() {
 	static get template() {
 		return html`
 		<style>
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -5,6 +5,7 @@ import '@brightspace-ui-labs/accordion/accordion.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import 'siren-entity/siren-entity.js';
 import '../localize-behavior.js';
+import '../mixins/d2l-sequences-completion-tracking-mixin.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 import { PolymerElement } from '@polymer/polymer/polymer-element.js';
@@ -15,6 +16,7 @@ import { PolymerElement } from '@polymer/polymer/polymer-element.js';
 
 class D2LSequenceLauncherUnit extends mixinBehaviors([
 	D2L.PolymerBehaviors.Siren.EntityBehavior,
+	D2L.Polymer.Mixins.Sequences.CompletionTrackingMixin,
 	D2L.PolymerBehaviors.Sequences.LocalizeBehavior
 ],
 PolymerElement
@@ -122,8 +124,12 @@ PolymerElement
 			dataAsvCssVars: String,
 			href: {
 				type: String,
+				observer: '_onHrefChanged',
 				reflectToAttribute: true,
 				notify: true
+			},
+			_previousEntity: {
+				type: String
 			},
 			role: {
 				type: String
@@ -164,7 +170,8 @@ PolymerElement
 
 	static get observers() {
 		return [
-			'_checkForEarlyLoadEvent(entity, subEntities)'
+			'_checkForEarlyLoadEvent(entity, subEntities)',
+			'_onEntityChanged(entity)'
 		];
 	}
 
@@ -176,6 +183,16 @@ PolymerElement
 		this.updateStyles(
 			styles
 		);
+	}
+
+	_onHrefChanged(href, previousHref) {
+		if (previousHref && previousHref !== href && this._previousEntity) {
+			this.finishPreviousEntityCompletion(this._previousEntity);
+		}
+	}
+
+	_onEntityChanged(entity) {
+		this._previousEntity = entity;
 	}
 
 	_getRootHref(entity) {
--- a/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
+++ b/d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
@@ -119,9 +119,6 @@ class D2LSequenceLauncherUnit extends D2L.Polymer.Mixins.Sequences.CompletionTra
 				reflectToAttribute: true,
 				notify: true
 			},
-			_previousEntity: {
-				type: String
-			},
 			role: {
 				type: String
 			},
@@ -161,8 +158,7 @@ class D2LSequenceLauncherUnit extends D2L.Polymer.Mixins.Sequences.CompletionTra
 
 	static get observers() {
 		return [
-			'_checkForEarlyLoadEvent(entity, subEntities)',
-			'_onEntityChanged(entity)'
+			'_checkForEarlyLoadEvent(entity, subEntities)'
 		];
 	}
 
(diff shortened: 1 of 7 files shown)


d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -1,24 +1,17 @@
...
 @mixes SirenEntityMixin
 */
 
-class D2LSequenceLauncherUnit extends mixinBehaviors([
-	D2L.PolymerBehaviors.Siren.EntityBehavior,
-	D2L.PolymerBehaviors.Sequences.LocalizeBehavior
-],
-PolymerElement
-) {
+class D2LSequenceLauncherUnit extends D2L.Polymer.Mixins.Sequences.CompletionTrackingMixin() {
```
imcdonald-d2l, line 14: Was not able to add this mixin to `mixinBehaviors`... not 100% sure why but I believe this shouldn't be an issue since `CompletionTrackingMixin()` extends `EntityBehavior` and `LocalizeBehavior`. Correct me if this is actually an issue though.

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -178,6 +176,16 @@ PolymerElement
 		);
 	}
 
+	_onHrefChanged(href, previousHref) {
+		if (previousHref && previousHref !== href && this._previousEntity) {
+			this.startPreviousEntityCompletion(this._previousEntity);
```
imcdonald-d2l, line 181: Added this function so completion information is fetched when navigating to a new entity in the nav-bar.

mixins/d2l-sequences-automatic-completion-tracking-mixin.js
```diff
@@ -59,7 +51,6 @@ function AutomaticCompletionTrackingMixin() {
 
 		_entityUpdated() {
 			if (this.href !== this._previousHref) {
-				this.finishCompletion();
 				this.startCompletion();
```
imcdonald-d2l, line 54: calling `startCompletion` here allows us to fetch completion information when navigating to a new item. This is particularly helpful when viewing an HTML topic because as soon as they navigate and view it, they have "completed" the activity. This is also useful in the case where they previously viewed and completed an activity but the  6 second delay did not accurately catch the completion information. Navigating back to the item at a different time will fetch the completion information and (hopefully) at that point the accurate completion is available.

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -83,74 +98,12 @@ function CompletionTrackingMixin() {
 			}
 		}
 
-		_fireToastEvent() {
```
imcdonald-d2l, line 86: The function isn't needed since we never know how true the completion information fetched is for particular activity items. The user may have indeed finished an activity, we just don't have the latest completion information so displaying a toast may be incorrect in some scenarios.

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -83,74 +98,12 @@ function CompletionTrackingMixin() {
...
-					composed: true,
-				});
-				window.dispatchEvent(event);
+		_isComplete(entity) {
+			const subEntity = entity.getSubEntityByClass('link-activity') || entity.getSubEntityByClass('file-activity');
+			if (!subEntity) {
+				return false;
 			}
-			this._failedCompletion = null;
+			return !!subEntity.getSubEntityByClass('completed');
```
imcdonald-d2l, line 106: `!!` simply returns the boolean value of the return value of the function. Same as: 
```
if (subEntity.getSubEntityByClass('complete')) {
   return true;
} else {
  return false;
}
```
More on the operation here: https://medium.com/better-programming/javascript-bang-bang-i-shot-you-down-use-of-double-bangs-in-javascript-7c9d94446054

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -18,36 +18,51 @@ function CompletionTrackingMixin() {
...
 				}
 			};
 		}
 
-		finishCompletion() {
-			this._fireToastEvent();
-
-			if (!this._completionEntity || this._skipCompletion) {
+		startCompletion() {
+			if (!this.entity || this._skipCompletion || this._isComplete(this.entity)) {
```
imcdonald-d2l, line 29: in either case of `startCompletion` or `startPreviousEntityCompletion` we can check completion status so we don't do unnecessary network calls to fetch completion information again. Unless activity items can revert back to an incomplete state?? I don't think that's the case but we should remove this condition if that's true.

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -18,36 +18,51 @@ function CompletionTrackingMixin() {
...
 
-			this._performViewActions(this._completionEntity, 'finish-view-activity')
-				.then(() => { this._completionEntity = null; })
-				.catch(() => { this._completionEntity = null; });
+			this._performViewActions(this.entity, 'view-activity-duration')
+				.then(completion => {
+					this._completionEntity = completion;
+					this._finishCompletion();
+				})
+				.catch(() => {});
```
ChrisKraljevicD2L, line 38: Catching and silencing errors doesn't sit right with me. This is how we can get things dying without any errors in the console.

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -18,36 +18,51 @@ function CompletionTrackingMixin() {
...
+			// will not appear complete if immediately fetching activity completion information
+			// after navigation to a new item
+			setTimeout(() => {
+				this._performViewActions(previousEntity, 'view-activity-duration')
+					.then(completion => {
+						this._completionEntity = completion;
+						this._finishCompletion();
+					})
+					.catch(() => {});
+			}, 6000);
```
ChrisKraljevicD2L, line 55: Where does the 6 second time interval come from?

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -18,36 +18,51 @@ function CompletionTrackingMixin() {
...
+			// need timeout as items such as quizes, assignments, discussions etc
+			// will not appear complete if immediately fetching activity completion information
+			// after navigation to a new item
+			setTimeout(() => {
+				this._performViewActions(previousEntity, 'view-activity-duration')
+					.then(completion => {
+						this._completionEntity = completion;
+						this._finishCompletion();
+					})
+					.catch(() => {});
```
ChrisKraljevicD2L, line 54: Same comment about silencing errors.

mixins/d2l-sequences-completion-tracking-mixin.js
imcdonald-d2l: Agreed, not a good approach. Should I just `console.error` the error? Not sure how errors are typically handled, or **should** be handled. I see a lot of silenced errors here and there but I'm assuming that's a result of the old team's poor development practices.

mixins/d2l-sequences-completion-tracking-mixin.js
imcdonald-d2l: Honestly, just a number out of a hat basically. Was testing locally and 6 second delay gives us a decent window to catch completion status of items that take longer to process (quizzes, topics, etc.)

mixins/d2l-sequences-completion-tracking-mixin.js
ChrisKraljevicD2L: Normally we can forego a `catch` statement and let our code explode unless we explicitly want to handle an error a certain way (with some sort of UI display normally). That way we, and our clients, can be sure that something is going wrong instead of guessing.

mixins/d2l-sequences-completion-tracking-mixin.js
ChrisKraljevicD2L: Alright, my concern with this is that we can't guarantee how good our clients' students' internet connections are. Some calls may take longer than 6 seconds on their networks. I doubt we could get Business Intelligence data on this but we could log some telemetry and adjust this later based on the data we get back.

Telemetry is definitely outside the scope of this PR though. I'm ok with 6 seconds for now but I'd like to get some data on how long this is taking as a followup.

imcdonald-d2l: I totally agree, it's not perfect to do a hard coded timeout but Ed and Brett were mentioning during stand up this morning these items are put on queues to get their completion status updated. So it is dependent on 

1. The user's internet connection
2. The status of the queue at any given moment. Could take longer to mark an item as complete during longer queues and less time during short queues.

But yes, telemetry may help us determine a better timeout. Ed and Brett also mentioned the idea of polling for completion status every 3 seconds (for example) until an item is marked as complete. This also has issues though and Ed and Brett didn't want to jump the gun on that solution just yet until they do further research into the state and importance of this issue.

mixins/d2l-sequences-completion-tracking-mixin.js
imcdonald-d2l: Ah, good to know. That's a fair point. I shall remove the empty catch statements! There shouldn't be any hidden issues there anyways.

mixins/d2l-sequences-completion-tracking-mixin.js
ChrisKraljevicD2L: Gotcha. Polling is a possible solution and I've made a library that could help with that. We previously implemented a polling solution for release conditions. At least we've got that in our back pocket if need be.

d2l-sequence-launcher-unit/d2l-sequence-launcher-unit.js
```diff
@@ -178,6 +176,16 @@ PolymerElement
...
+		}
+	}
+
+	_onEntityChanged(entity) {
+		this._previousEntity = entity;
```
ianlaidlaw, line 186: i'm not sure i understand this. wouldn't the parameter `entity` be the most up to date entity, so not the previous one?

ianlaidlaw: okay i think i see what's going on

imcdonald-d2l: Good catch, you are actually correct in a sense here. `_previousEntity` is just being assigned to the most recent entity. But since the `_onHrefChanged()` function is triggered before an entity update this logic works. So all we actually need is this: 
```
_onHrefChanged(href, previousHref) {
	// this.entity is the previous entity on href change
	// since it has not been updated at this point
	if (previousHref && previousHref !== href && this.entity) {
			this.startPreviousEntityCompletion(this.entity);
	}
}
```

ianlaidlaw: Okay, I think that will work. To me it looks like we can guarantee that `entity` will not be updated yet because it's dependent on `href`, so `_onHrefChanged` should get called before entity gets changed.

imcdonald-d2l: That is correct, entity would be dependent on `href` so it should be pretty safe to assume that.

mixins/d2l-sequences-completion-tracking-mixin.js
ianlaidlaw: since this 6 seconds could change or may be misleading, we could put it into a constant somewhere so this isn't super hidden. Like at the top of the file have something like:

```
const COMPLETION_TRACKING_DELAY = 6000;
```

imcdonald-d2l: Agreed, that is a good idea @ianlaidlaw 👍

pbrown-d2l:
One thing to test before merging this is to verify that viewing time is still being tracked correctly. You should be able to see a user's time viewing a topic in the User Progress tool.

imcdonald-d2l:
> One thing to test before merging this is to verify that viewing time is still being tracked correctly. You should be able to see a user's time viewing a topic in the User Progress tool.

@pbrown-d2l good catch, did some testing and this feature is not working as expected now. Will have to tweak some stuff to see if I can get it working again.. I have an idea that may work.

PR #248, opened 2020-07-09
Kill d2l-sequence-viewer

ChrisKraljevicD2L:
I'm open to arguments for whether or not this warrants a major version bump. I personally think so, but I can be swayed.

Once this is merged I'll make a BSI PR to use the new `sequences` `d2l-sequence-viewer` component.

(I've also tested with BSI locally to make sure things are a-ok)


CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Team Phoenix Members
+
+* @ChrisKraljevicD2L @ianlaidlaw @sarahboyd @pbrown-d2l
```
ChrisKraljevicD2L, line 3: I'm not sure when our `CODEOWNERS` file should be expanded but this is copied from `d2l-sequence-viewer` for now.

demo/d2l-sequence-viewer/data/dummy
```diff
@@ -0,0 +1 @@
+dummy text
```
ianlaidlaw, line 1: what is this for?

CODEOWNERS
apalaniuk: You can also do `@Phoenix` and not have to manage these per-repo if that fits your needs.

DiljotSG:
https://www.youtube.com/watch?v=i7gIpuIVE3k

demo/d2l-sequence-viewer/data/dummy
ChrisKraljevicD2L: I have a feeling it was once for [this activity](https://github.com/BrightspaceHypermediaComponents/sequences/blob/ckraljevic/kill-sequence-viewer/demo/d2l-sequence-viewer/data/l1-m2-activity7.json#L44) but I don't see any `href`s that refer to the actual `dummy.txt` file. I'll nuke it.

ChrisKraljevicD2L:
@ianlaidlaw I agree that directories can definitely be squished. Possibly as a followup :+1:

CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# Team Phoenix Members
+
+* @Phoenix
```
apalaniuk, line 3: Think I flubbed that shooting from the hip, that should be `@Brightspace/Phoenix` since these are global identifiers

PR #250, opened 2020-07-15
Clean up completion mixin and add delay

imcdonald-d2l:
Cleaning up unnecessary bits of completion mixin as well as adding a timeout so fetching completion information is delayed when navigating off an activity item. The purpose of this timeout is to give us a larger time window for the completion queue to mark activity items as complete before fetching them. The timeout duration is currently set at 6 seconds but can easily be tweaked later if we find a better delay is more suitable.

### DIFF
--- a/mixins/d2l-sequences-completion-tracking-mixin.js
+++ b/mixins/d2l-sequences-completion-tracking-mixin.js
@@ -5,6 +5,9 @@ import '../localize-behavior.js';
 import { Maybe } from '../maybe.js';
 import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 import { PolymerElement } from '@polymer/polymer/polymer-element.js';
+
+const FINISH_COMPLETION_DELAY = 6000;
+
 function CompletionTrackingMixin() {
 	return class extends mixinBehaviors([
 		D2L.PolymerBehaviors.Siren.EntityBehavior,
@@ -18,9 +21,6 @@ function CompletionTrackingMixin() {
 				_completionEntity: {
 					type: Object
 				},
-				_failedCompletion: {
-					type: Object
-				},
 				_skipCompletion: {
 					type: Boolean,
 					computed: '_isImpersonating(token)'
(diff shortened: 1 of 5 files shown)


mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -82,76 +80,6 @@ function CompletionTrackingMixin() {
 				return false;
 			}
 		}
-
-		_fireToastEvent() {
```
imcdonald-d2l, line 86: toast is unnecessary as completion status is not very accurate when fetched

test/d2l-sequences-completion-tracking-mixin.html
```diff
@@ -99,18 +99,6 @@
 		expect(performViewActions).not.to.have.been.called;
 	});
 
-	it('finishCompletion must perform siren action: finish-view-activity', async() => {
```
imcdonald-d2l, line 102: Can't test this when it's wrapped in a `setTimeout()`

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -29,15 +29,14 @@ function CompletionTrackingMixin() {
...
 				return;
 			}
 
-			this._performViewActions(this._completionEntity, 'finish-view-activity')
-				.then(() => { this._completionEntity = null; })
-				.catch(() => { this._completionEntity = null; });
+			setTimeout(() => {
+				this._performViewActions(this._completionEntity, 'finish-view-activity')
+					.then(() => { this._completionEntity = null; });
+			}, FINISH_COMPLETION_DELAY);
```
prajapac, line 39: What does the delay achieve? Also, how was the delay value of 6s determined?

imcdonald-d2l: After completing activity items such as quizzes and discussions they are put on queues in the system to be marked "complete".  Since theses queues take time, fetching completion information right after navigating off the activity item results in the item being seen as incomplete since the item hasn't made it through the queue at that point in time. This delay is allowing us to give the queue some time to process the item before fetching the completion information. ... as for the 6 seconds specifically, that was a number I chose based off testing on my local instance. It is by no means a number based off data or hard evidence but a temporary patch to increase the likelihood of fetching completion status correctly. This number will likely change in the future after we obtain some sort of data on how long a typical queue delay is.

PR #252, opened 2020-07-16
US118394 revert the revert - re-update d2l-video

sarahboyd:
We want to ship the updated d2l-video component in the **September** release (not August). This should be merged **after 20.20.8 branching**.

Re-adds back in https://github.com/BrightspaceHypermediaComponents/sequences/pull/245

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@brightspace-ui-labs/accordion": "^2.3.0",
     "@brightspace-ui/core": "^1",
     "@d2l/audio": "Brightspace/d2l-audio#semver:^3",
-    "@d2l/video": "Brightspace/d2l-video#semver:^1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "@brightspace-ui-labs/accordion": "^2.3.0",
     "@brightspace-ui/core": "^1",
     "@d2l/audio": "Brightspace/d2l-audio#semver:^3",
-    "@d2l/video": "Brightspace/d2l-video#semver:^1",
+    "@d2l/video": "Brightspace/d2l-video#semver:^2",
```
prajapac, line 70: What's the differences between the different versions?

sarahboyd: a few style changes but the big change is a new playback speed control button. We want to actually announce the new feature in the release notes, that's why we're delaying a month

PR #253, opened 2020-07-21
Cprajapati/de39648

prajapac:
This PR includes changes that make images render responsively in LX. This was in response to: [DE39648](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F400766093372), where larger images led to scrollbars being shown and the images not being rendered within the view window. 

This was because the images were being rendered in an iframe.. 😞 

The new component, D2LSequencesContentImage, simply loads the image in a div and using the img tag.

Before:
Before

After:
After

### DIFF
--- a/components/d2l-sequences-content-image.js
+++ b/components/d2l-sequences-content-image.js
@@ -1,0 +1,56 @@
+import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
+	static get template() {
+		return html`
+		<style>
+		    img {
+                max-width: 100%;
+                height: auto;
+			}
+		</style>
+		<div>
+            <img src="[[_fileLocation]]">
+		</div>
+`;
+	}
+
+	static get is() {
+		return 'd2l-sequences-content-image';
+	}
+	static get properties() {
+		return {
+			href: {
+				type: String,
+				reflectToAttribute: true,
+				notify: true,
+				observer: '_scrollToTop'
+			},
+			_fileLocation: {
+				type: String,
+				computed: '_getFileLocation(entity)'
+			},
+			title: {
+				type: String,
+				computed: '_getTitle(entity)'
+			},
+		};
+	}
+	_scrollToTop() {
+		window.top.scrollTo(0, 0);
+	}
+	_getFileLocation(entity) {
+		try {
+			const fileActivity = entity.getSubEntityByClass('file-activity');
+			const file = fileActivity.getSubEntityByClass('file');
+			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			return link.href;
+		} catch (e) {
+			return '';
+		}
+	}
+	_getTitle(entity) {
+		return entity && entity.properties && entity.properties.title || '';
+	}
+}
+customElements.define(D2LSequencesContentImage.is, D2LSequencesContentImage);
--- a/components/d2l-sequences-content-image.js
+++ b/components/d2l-sequences-content-image.js
@@ -1,5 +1,6 @@
 import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+
 export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
 	static get template() {
 		return html`
@@ -9,9 +10,7 @@ export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.Autom
                 height: auto;
 			}
 		</style>
-		<div>
-            <img src="[[_fileLocation]]">
-		</div>
+		<img src="[[_fileLocation]]">
 `;
 	}
 
@@ -29,28 +28,21 @@ export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.Autom
 			_fileLocation: {
 				type: String,
 				computed: '_getFileLocation(entity)'
-			},
-			title: {
-				type: String,
-				computed: '_getTitle(entity)'
-			},
+			}
 		};
 	}
 	_scrollToTop() {
 		window.top.scrollTo(0, 0);
 	}
 	_getFileLocation(entity) {
-		try {
+		if (entity) {
 			const fileActivity = entity.getSubEntityByClass('file-activity');
 			const file = fileActivity.getSubEntityByClass('file');
-			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			const link = file.getLinkByClass('embed') || file.getLinkByRel('alternate');
 			return link.href;
-		} catch (e) {
+		} else {
 			return '';
 		}
-	}
-	_getTitle(entity) {
-		return entity && entity.properties && entity.properties.title || '';
 	}
 }
 customElements.define(D2LSequencesContentImage.is, D2LSequencesContentImage);
--- a/components/d2l-sequences-content-router.js
+++ b/components/d2l-sequences-content-router.js
@@ -5,6 +5,7 @@ import { D2LSequencesContentFileHtml } from './d2l-sequences-content-file-html.j
 import { D2LSequencesContentFilePdf } from './d2l-sequences-content-file-pdf.js';
 import { D2LSequencesContentVideo } from './d2l-sequences-content-video';
 import { D2LSequencesContentAudio } from './d2l-sequences-content-audio';
+import { D2LSequencesContentImage } from './d2l-sequences-content-image';
 import './d2l-sequences-content-file-html.js';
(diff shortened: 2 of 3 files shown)


components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
+import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
+	static get template() {
+		return html`
+		<style>
+		    img {
+                max-width: 100%;
```
prajapac, line 8: This makes it so larger images will scale down for smaller displays, but will remain its original size for larger displays. To change it I would simply have to make it be width instead of max-width.

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+		</div>
+`;
+	}
+
+	static get is() {
+		return 'd2l-sequences-content-image';
+	}
+	static get properties() {
+		return {
+			href: {
```
imcdonald-d2l, line 23: The `href` and `title` property don't seem to be used in this component and should be removed if they aren't needed. Also, if `title` property is removed, `_getTitle` should be removed as well.

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+	_getFileLocation(entity) {
+		try {
+			const fileActivity = entity.getSubEntityByClass('file-activity');
+			const file = fileActivity.getSubEntityByClass('file');
+			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			return link.href;
+		} catch (e) {
+			return '';
+		}
```
imcdonald-d2l, line 50: Probably shouldn't have an empty catch statement. Was debugging and looks like this function is hit twice, once when entity is null and another time when the entity is populated so it could probably just be written like this instead: 

```
if (entity) {
   const fileActivity = entity.getSubEntityByClass('file-activity');
   const link = file.getLinkByClass('embed') || file.getLinkByRel('alternate');
   return link.href;
} else {
  return '';
}
```
Note that we don't need `file.getLinkByClass('pdf')` either since we know it won't be a pdf file.

components/d2l-sequences-content-image.js
imcdonald-d2l: Scaling down images makes sense to me so mobile users have a better experience 👍

components/d2l-sequences-content-image.js
ianlaidlaw: If I'm looking at this right this component is using `D2L.PolymerBehaviors.Siren.EntityBehavior` because it extends the `AutomaticCompletionTrackingMixin` which eventually uses that up the chain so `href` would be used in this component. I think sometimes we omit it because it's defined in that mixin. But yeah it looks like `title` isn't used at all and can be deleted.

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+		};
+	}
+	_scrollToTop() {
+		window.top.scrollTo(0, 0);
+	}
+	_getFileLocation(entity) {
+		try {
+			const fileActivity = entity.getSubEntityByClass('file-activity');
+			const file = fileActivity.getSubEntityByClass('file');
+			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
```
ianlaidlaw, line 46: If I understand this pr we're using this new component instead of `d2l-sequences-content-file.js` which renders an iframe (and why the image isn't showing properly). Do we need to check if the class is `pdf` ? I think we wouldn't want to show pdfs in this component

components/d2l-sequences-content-image.js
imcdonald-d2l: Good catch, so `href` is passed into `AutomaticCompletionTrackingMixin` from this component @ianlaidlaw? Just clarifying so I understand how the `href` property is working. Hard to wrap my head around that one since it doesn't seem to have a clear origin.  (or I can't find it)

components/d2l-sequences-content-image.js
prajapac: > Scaling down images makes sense to me so mobile users have a better experience 👍

How about the other way around, should small images be scaled up?

components/d2l-sequences-content-image.js
ianlaidlaw: Yeah so if you follow the chain...
This component extends `AutomaticCompletionTrackingMixin` which extends `CompletionTrackingMixin` which extends `D2L.PolymerBehaviors.Siren.EntityBehavior` which has an `href` property... So that means this component here is expecting an `href` and is how it gets populated with `entity`.

It's definitely not great and is really confusing how all this works together... But it is what it is.

ianlaidlaw: In most components if it just needs to get an entity you can omit writing it in this component, but if you need to do other things with it like add observers (the scroll to top here) you'll need to redeclare it.

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
+	static get template() {
+		return html`
+		<style>
+		    img {
+                max-width: 100%;
+                height: auto;
+			}
+		</style>
+		<div>
```
ianlaidlaw, line 12: small nit: is this div needed?

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
+import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
+import { html } from '@polymer/polymer/lib/utils/html-tag.js';
+export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
```
ianlaidlaw, line 3: nit: generally we leave a blank line between imports and everything else

components/d2l-sequences-content-image.js
ianlaidlaw: Sorry I missed Ian already commented about this!

components/d2l-sequences-content-image.js
prajapac: Not really, I suppose I could get rid of it. If I understand correctly, this part of LX just loads one resource at a time, i.e. the image only in this case. I tested it with and without the div and both worked fine.

components/d2l-sequences-content-image.js
prajapac: Ok, So I totally did that, but all the other files: d2l-sequences-content{video, audio, file-pdf, etc.}.js have the same and I just wanted to keep things consistent.

components/d2l-sequences-content-image.js
prajapac: I will remove the title property. The reason I added these two is since all other files have it (even when they don't seemingly use those properties): d2l-sequences-content{video, audio, file-pdf, etc.}.js

components/d2l-sequences-content-image.js
prajapac: Will do. (Again just saw all the other files had it so added it for consistency: d2l-sequences-content{video, audio, file-pdf, etc.}.js)

components/d2l-sequences-content-image.js
ChrisKraljevicD2L: Additional nit: our other file rendering components start with `D2LSequencesContentFile`, and I think that should continue here. That's just my opinion though.

prajapac: The audio and video components are named: D2LSequencesContentAudio and D2LSequencesContentVideo (i.e. no file in component name), so I followed suit with: D2LSequencesContentImage. 

On the other hand, file is appened to the component names for pdf, html and download: D2LSequencesContentFilePdf, D2LSequencesContentFileHtml and D2LSequencesContentFileDownload.

ChrisKraljevicD2L: Seems like it's less consistent than I thought. Carry on!

ianlaidlaw: It's also worth noting that these components weren't written by us and probably don't meet our standards - no harm improving things where we can

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+				type: String,
+				reflectToAttribute: true,
+				notify: true,
+				observer: '_scrollToTop'
+			},
+			_fileLocation: {
+				type: String,
+				computed: '_getFileLocation(entity)'
+			},
+			title: {
```
ianlaidlaw: this is unused, can delete

components/d2l-sequences-content-image.js
```diff
@@ -0,0 +1,56 @@
...
+		try {
+			const fileActivity = entity.getSubEntityByClass('file-activity');
+			const file = fileActivity.getSubEntityByClass('file');
+			const link = file.getLinkByClass('pdf') || file.getLinkByClass('embed') || file.getLinkByRel('alternate');
+			return link.href;
+		} catch (e) {
+			return '';
+		}
+	}
+	_getTitle(entity) {
```
ianlaidlaw: this is unused, can delete

PR #258, opened 2020-07-29
DE39748: Make (large) image responsively fit screen, i.e. vertically as well i…

prajapac:
The previous change I had made, made it so large images would scale down to fit horizontally but not vertically. The expected case is for the image to scale down vertically as well.

This has been achieved by making the max-height property to also have the value of 100%, in addition, margin: auto makes the image center horizontally and display: block makes it so the image fits and no scroll bars are visible.

Before:
Before-01

After:
After-01

### DIFF
--- a/components/d2l-sequences-content-image.js
+++ b/components/d2l-sequences-content-image.js
@@ -6,8 +6,10 @@ export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.Autom
 		return html`
 		<style>
 		    img {
+				display: block;
+				margin: auto;
                 max-width: 100%;
-                height: auto;
+                max-height: 100%;
 			}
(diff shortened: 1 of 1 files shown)


sarahboyd:
Looks great!
>display: block makes it so the image fits and no scroll bars are visible
this will make people super happy!

components/d2l-sequences-content-image.js
```diff
@@ -6,8 +6,10 @@ export class D2LSequencesContentImage extends D2L.Polymer.Mixins.Sequences.Autom
 		return html`
 		<style>
 		    img {
+				display: block;
+				margin: auto;
```
imcdonald-d2l, line 10: nit: the tab indentation should be consistent with the properties below.

PR #263, opened 2020-08-05
Add support for deferred doc conversion

ianlaidlaw:
This pr adds support for deferred Doc Conversion, added in this pr: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/14462/overview

The problem was that sometimes conversion would take a long time and time out, breaking LX. In order for this to work, I had to add support for the "processing" class, which previously existed but was not used at all.

The way this works for a deferred conversion is as follows:
- LX receives "processing" class and knows to render the "processing" component
- The processing component has polling support to continuously refetch the entity, bypassing the entity store cache
- Every poll, the d2l-sequences-content-router will check which component to render. If conversion is not done, it will still have the processing class. If it has the converted doc, then it will render that one.
- Since LX previously prevented rerendering when the href has changed, I added logic to detect when transitioning off the processing component.

View of the processing component:

image


components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,89 @@
...
+		return {
+			href: {
+				type: String,
+				reflectToAttribute: true,
+				notify: true,
+			},
+			_poller: Object,
+			_pollIncrement: {
+				type: Number,
+				value: 2000
```
ianlaidlaw, line 39: I pulled these numbers out of thin air. Confirm with product or Grant?

components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,89 @@
...
+*/
+export class D2LSequencesContentFileProcessing extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
+	static get template() {
+		return html`
+		<style>
+			:host {
+				display: flex;
+				height: 100%;
+			}
+			d2l-loading-spinner {
```
ianlaidlaw, line 18: There was no design for this so I just made it a centered loading spinner. Confirm with design

components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,89 @@
...
+
+	async _poll() {
+		await this._refreshEntity();
+		if (this._pollInterval < this._pollMax) {
+			this._pollInterval += this._pollIncrement;
+			this._poller.setupPolling(this._pollInterval);
+		}
+	}
+
+	async _refreshEntity() {
```
ianlaidlaw, line 81: This looks so bad.. but it seems to work. I just "rehit" the same href and it goes through and refetches it

mixins/d2l-sequences-router-mixin.js
```diff
@@ -71,9 +73,12 @@ function RouterMixin(getEntityType) {
 						? D2LSequencesContentError.is
 						: getEntityType(entity);
 
+					const fileWasProcessing = this._contentType === D2LSequencesContentFileProcessing.is &&
```
ianlaidlaw, line 76: I don't love this name. But I need a way to refresh the view when transitioning off of a FileProcessing view. Open to suggestions

mixins/d2l-sequences-router-mixin.js
```diff
@@ -71,9 +73,12 @@ function RouterMixin(getEntityType) {
...
+						entityType !== D2LSequencesContentFileProcessing.is;
+
 					const replaceContentElement = (entity &&
```
ianlaidlaw, line 79: bonus points to someone who can rewrite this to be less ugly

components/d2l-sequences-content-router.js
```diff
@@ -100,7 +105,11 @@ function getFileEntityType(fileActivity) {
 	const file = fileActivity.getSubEntityByClass('file');
 	let mimeType = file && file.properties && file.properties.type;
 
-	if (file.getLinkByClass('d2l-converted-doc')) {
+	// A converted doc will either be processing,
+	// or present with the d2l-converted-doc class
+	if (fileActivity.getSubEntityByClass('processing')) {
+		return D2LSequencesContentRouter.fileProcessing;
+	} else if (file.getLinkByClass('d2l-converted-doc')) {
```
ianlaidlaw, line 112: ideally this line shouldn't really make this assumption yet... It could have failed (we don't have a view for that, I think it would just go to the download page...). We should be checking the other statuses on the `fileActivity` before jumping down here. But since this is "working" I'm gonna leave this for now..

components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,89 @@
...
+		// Override the entity cached in EntityStore
+		await window.D2L.Siren.EntityStore.fetch(this.href, this.token, true);
+		const that = this;
+		this.href = that.href;
```
imcdonald-d2l, line 85: Do mind explaining what this is doing? A bit confused on what 84/85 is doing 🤔

ianlaidlaw: Basically I am "rehitting" the same href in order to refetch the entity as well as make it go through the same steps the component goes through when it gets an entity, specifically determining the content type, etc.

All I'm trying to do is assign `this.href` to a different variable but it's the same value. I don't think if I did:
`this.href = this.href` it would work, but I actually didn't try.

Alternatively I could do:
```
const refreshedEntity = await window.D2L.Siren.EntityStore.fetch(this.href, this.token, true);
this.href = refreshedEntity.href; // (same value)
```

imcdonald-d2l: Ah that makes more sense now, thanks! The alternate suggestion you made may be a better approach so it isn't as confusing to others maybe?

components/d2l-sequences-content-file-processing.js
pbrown-d2l: I don't think there is an easier way to do this right now

components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,94 @@
...
+			}
+			.loading-container {
+				display: flex;
+				flex-direction: column;
+				margin: auto;
+			}
+		</style>
+		<div class="loading-container">
+			<d2l-loading-spinner size="50"></d2l-loading-spinner>
+			<span>Converting Document</span>
```
ianlaidlaw, line 27: Oops, I created the lang terms but forgot to use them. Will update.

components/d2l-sequences-content-file-processing.js
```diff
@@ -0,0 +1,98 @@
...
+				display: flex;
+				height: 100%;
+			}
+			.loading-container {
+				display: flex;
+				flex-direction: column;
+				margin: auto;
+			}
+		</style>
+		<div class="loading-container">
```
ianlaidlaw, line 29: image

got this from design

lang/ar-AE.json
```diff
@@ -10,7 +10,8 @@
     "congratulations": "تهانينا!",
     "contentLinkText": "انقر للانتقال إلى المحتوى الجديد",
     "continue": "متابعة",
-    "download": "تنزيل",
+	"convertingDocument": "Converting Document",
+	"download": "تنزيل",
```
ianlaidlaw, line 14: The reason why some of the below lines for "download" changed is because some of the lang files don't use a tab, they are 3 spaces away... and my IDE automatically sets it to a 4 space tab as is standard for the rest of the project

components/d2l-sequences-content-file-processing.js
ianlaidlaw: I got rid of `that` and get the href off the refreshed entity. hopefully is more clear

components/d2l-sequences-content-file-processing.js
```diff
@@ -79,11 +85,10 @@ export class D2LSequencesContentFileProcessing extends D2L.Polymer.Mixins.Sequen
 	}
 
 	async _refreshEntity() {
-		// Override the entity cached in EntityStore
-		await window.D2L.Siren.EntityStore.fetch(this.href, this.token, true);
-		const that = this;
-		this.href = that.href;
+		// Hit the same href to bypass and override the entity cached in EntityStore
+		const { entity: refreshedEntity } = await window.D2L.Siren.EntityStore.fetch(this.href, this.token, true);
+		this.href = refreshedEntity.getLinkByRel('self').href;
```
imcdonald-d2l, line 94: do we need error catching here or is it safe to assume we will always get a refreshed entity?

ianlaidlaw: hmm that's a good question. I suppose it could be possible that the course creator could delete the topic _while_ the user is sitting on the processing screen. In that case this would error out. I'm not sure what we expect to happen in that case though. I'll try and test this when I get this running on an external site

ianlaidlaw: At the very least I could just not set the href if it doesn't exist, then it would just spin forever lol.

pbrown-d2l: I think an error in that case is not bad. It would happen if a user clicked on a TOC entry to a topic that was just deleted

pbrown-d2l: It is also very unlikely to happen in the wild

imcdonald-d2l: Yeah, if there is a predictable way to handle errors that would be good. As long as LX doesn't crap out we're good. Infinite loading would be better than blowing up probably.

ianlaidlaw: Yeah, that's why I wasn't too concerned about it. It's so unlikely to happen and would be resolved by a simple refresh. and we have so many other situations where if the creator deleted a topic it would break LX, not just this. I think if it sits on the Loading spinner forever, that's fine

PR #267, opened 2020-08-25
Move scroll bar on HTML activities to far right

imcdonald-d2l:
Adjusted sequence viewer styles so that HTML activities had the scroll bar on the far right. Had to inject CSS styles on `<body>` tag in `<iframe>` since that is where the scroll bar lives. Should be a relatively safe change since all html content will be injected in the `<body>` tag and I assume this won't change over time. Attached below are screenshots with the new scroll bar and visuals on what the injected styles look like. Open to suggestions for improvements or if you think this change is worth it.

Since our team does not own other activities like discussions, assignments, checklists, etc. it is difficult to manipulate the DOM styling similarly to the HTML component since their DOM structure could change without us knowing causing any changes to break in the future. 

html-without-nav
html-with-nav

### DIFF
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -1,24 +1,31 @@
 import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
+import { VIEWER_MAX_WIDTH, VIEWER_HORIZONTAL_MARGIN } from '../util/constants';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
 	static get template() {
 		return html`
-		<style>
-			.d2l-sequences-scroll-container {
-				-webkit-overflow-scrolling: touch;
-				overflow-y: auto;
-				height: 100%;
-			}
-			iframe {
-				width: 100%;
-				height: calc(100% - 12px);
-				overflow: hidden;
-			}
-		</style>
-		<div class="d2l-sequences-scroll-container">
-			<iframe id="content" frameborder="0" src$="[[_fileLocation]]" title$="[[title]]" allowfullscreen allow="microphone *; camera *; display-capture *; encrypted-media *;"></iframe>
-		</div>
-`;
+			<style>
+				.d2l-sequences-content-container {
+					height: 100%;
+				}
+				iframe {
+					width: 100%;
+					height: calc(100% - 12px);
+					overflow: hidden;
+				}
+			</style>
+			<div class="d2l-sequences-content-container">
+				<iframe
+					id="content"
+					on-load="_setIframeStyles"
+					frameborder="0"
+					src$="[[_fileLocation]]"
+					title$="[[title]]"
+					allowfullscreen
+					allow="microphone *; camera *; display-capture *; encrypted-media *;"
+				></iframe>
+			</div>
+		`;
 	}
 
 	static get is() {
@@ -57,6 +64,13 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 		super.disconnectedCallback();
 		window.removeEventListener('d2l-sequence-viewer-multipage-navigation', this._navigateMultiPageFileListener);
 		window.postMessage(JSON.stringify({ handler: 'd2l.nav.reset' }), '*');
+	}
+
+	_setIframeStyles() {
+		const htmlIframe = this.$.content;
+		const maxWidth = VIEWER_MAX_WIDTH + (2 * VIEWER_HORIZONTAL_MARGIN);
+		htmlIframe.contentDocument.body.style.maxWidth = `${maxWidth}px`;
+		htmlIframe.contentDocument.body.style.margin = '0 auto';
 	}
 
 	_scrollToTop() {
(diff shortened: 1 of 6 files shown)


components/d2l-sequences-content-file-html.js
```diff
@@ -1,24 +1,31 @@
 import '../mixins/d2l-sequences-automatic-completion-tracking-mixin.js';
+import { VIEWER_MAX_WIDTH, VIEWER_HORIZONTAL_MARGIN } from '../util/constants';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.AutomaticCompletionTrackingMixin() {
 	static get template() {
 		return html`
-		<style>
-			.d2l-sequences-scroll-container {
-				-webkit-overflow-scrolling: touch;
-				overflow-y: auto;
```
imcdonald-d2l, line 9: These actually did nothing since the scroll bar existed in the `iframe` and not on the container.

components/d2l-sequences-content-file-html.js
```diff
@@ -1,24 +1,31 @@
...
+				iframe {
+					width: 100%;
+					height: calc(100% - 12px);
+					overflow: hidden;
+				}
+			</style>
+			<div class="d2l-sequences-content-container">
+				<iframe
+					id="content"
+					on-load="_setIframeStyles"
```
imcdonald-d2l, line 20: Have to set styles `on-load` or it won't work properly.

d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -117,10 +123,9 @@ class D2LSequenceViewer extends mixinBehaviors([
 					display: inline-block;
 					height: 100%;
 					overflow-y: auto;
-					margin: 0 var(--viewframe-horizontal-margin);
 				}
-				#viewframe:focus {
-					outline: none;
+				:host([is-not-html-entity]) .viewer {
+					margin: 0 var(--viewframe-horizontal-margin);
```
imcdonald-d2l, line 128: replaced by `padding` on `<body>` tag in HTML component.

d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -96,18 +98,22 @@ class D2LSequenceViewer extends mixinBehaviors([
...
 					flex: 2;
-					margin: 0 auto;
 					padding: 18px 0;
 					flex-direction: column;
 				}
+				:host([is-not-html-entity]) #viewframe {
+					/*Viewframe max width is 1170px, but viewer has 30px
+					inherent padding horizontally to account for.*/
+					max-width: calc(var(--viewer-max-width) + 2*var(--viewframe-horizontal-margin));
+					margin: 0 auto;
```
imcdonald-d2l, line 112: replaced by `max-width` and `margin` properties on `<body>` tag in HTML component.

util/constants.js
```diff
@@ -0,0 +1,2 @@
+export const VIEWER_MAX_WIDTH = 1170;
+export const VIEWER_HORIZONTAL_MARGIN = 30;
```
imcdonald-d2l, line 2: created this file since I thought other components might use these as well but ultimately found that the HTML activity is the only relatively safe component to adjust for scrolling. Can move these constants to the html component if we think that's best.

components/d2l-sequences-content-file-html.js
ianlaidlaw: We might need to double check that this works on iOS

components/d2l-sequences-content-router.js
```diff
@@ -63,7 +63,7 @@ class D2LSequencesContentRouter extends D2L.Polymer.Mixins.Sequences.RouterMixin
 }
 customElements.define(D2LSequencesContentRouter.is, D2LSequencesContentRouter);
 
-function getEntityType(entity) {
+export function getEntityType(entity) {
```
ianlaidlaw, line 66: exporting this from this class doesn't feel quite right. if you need to reuse this logic then perhaps it should be extracted to some helper or a mixin or something. the logic associated with this component should be encapsulated to this file, imo. open to ideas here

ianlaidlaw: things like this leads to circular dependencies

imcdonald-d2l: hmmm I get what you are saying but not sure if it is necessary for this file. If you look at the contents of the file it is basically just a utility to get the content type and return the component. It's already pretty bare bones and `getEntityType` basically uses every function defined in that component so not sure if we need to extract it out. If we were to extract it out to a new file we would have to copy basically everything in the component file with it. Essentially just duplicating the component.

imcdonald-d2l: Right now this component is only used in `d2l-sequence-viewer` as well, so pretty safe _I think_.

components/d2l-sequences-content-file-html.js
imcdonald-d2l: I can keep them in to be safe if we are concerned? Doesn't hurt having them there, they won't wreck anything

d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -18,8 +18,10 @@ import '@brightspace-ui/core/components/button/button-subtle.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 import { PolymerElement } from '@polymer/polymer/polymer-element.js';
+import { getEntityType } from '../components/d2l-sequences-content-router.js';
```
ChrisKraljevicD2L, line 21: It looks pretty sus to import a function from a component. I agree with Ian that this should be in a util/helper library.

components/d2l-sequences-content-file-html.js
pbrown-d2l: IOS just hates

PR #270, opened 2020-09-15
Prevent contentDocument error and style accordingly

imcdonald-d2l:
When the parent and iframe domain are different, `contentDocument` returns `null` and we can't properly style the contents of the iframe. To prevent a JavaScript error when `contentDocument` is `null` we only style the iframe when we know `contentDocument` (and related properties) exist. 

To ensure that we style the html document consistently regardless of the existence of `contentDocument` I have added back-up styles. These back-up styles ensure the proper width and margin are set on the iframe while sacrificing the ability to have the scroll bar on the far right. Without these back-up styles the html document would take up the full width of the screen and lack consistency with the other activity styles in LX.

**Desired styles** (domains match and we can style the iframe with the scroll bar on the far right): 
scroll-bar-right

**Back-up styles** (see that the scroll bar is directly beside the html content):
scroll-bar

### DIFF
--- a/components/d2l-sequences-content-file-html.js
+++ b/components/d2l-sequences-content-file-html.js
@@ -69,11 +69,16 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 	}
 
 	_setIframeStyles() {
+		// If parent domain does not match fileLocation domain, contentDocument will return null
+		// and log an error in the console. Code is wrapped in if-statement to clean up console errors
+		// and prevent unnecessary errors
 		const htmlIframe = this.$.content;
 		const maxWidth = VIEWER_MAX_WIDTH + (2 * VIEWER_HORIZONTAL_MARGIN);
-		htmlIframe.contentDocument.body.style.maxWidth = `${maxWidth}px`;
-		htmlIframe.contentDocument.body.style.margin = '0 auto';
-		htmlIframe.contentDocument.body.style.padding = `0 ${VIEWER_HORIZONTAL_MARGIN}px`;
+		if (htmlIframe && htmlIframe.contentDocument && htmlIframe.contentDocument.body) {
+			htmlIframe.contentDocument.body.style.maxWidth = `${maxWidth}px`;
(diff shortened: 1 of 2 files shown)


CodeBaboon (approved):

components/d2l-sequences-content-file-html.js
```diff
@@ -70,10 +91,16 @@ export class D2LSequencesContentFileHtml extends D2L.Polymer.Mixins.Sequences.Au
 
 	_setIframeStyles() {
 		const htmlIframe = this.$.content;
-		const maxWidth = VIEWER_MAX_WIDTH + (2 * VIEWER_HORIZONTAL_MARGIN);
-		htmlIframe.contentDocument.body.style.maxWidth = `${maxWidth}px`;
-		htmlIframe.contentDocument.body.style.margin = '0 auto';
-		htmlIframe.contentDocument.body.style.padding = `0 ${VIEWER_HORIZONTAL_MARGIN}px`;
+		if (htmlIframe && htmlIframe.contentDocument && htmlIframe.contentDocument.body) {
```
dlockhart, line 94: I still have a feeling simply accessing `htmlIFrame.contentDocument` or `contentDocument.body` is going to throw, so your code isn't going to be reached. (if you tested it though and it does, that's cool)

Typically we do something like this:
```javascript
try {
    htmlIframe.contentDocument.body.style.maxWidth = `${MAX_WIDTH}px`;
    htmlIframe.contentDocument.body.style.margin = '0 auto';
    htmlIframe.contentDocument.body.style.padding = `0 ${VIEWER_HORIZONTAL_MARGIN}px`;
} catch( e ) {
    this.backUpStyles = true;
}
```

imcdonald-d2l: `contentDocument` will exist as a property, it will just be set to `null` so I don't expect this to error out. The `else` statement will simply execute when it is `null`. Can't test in a production state locally but am fairly confident this won't throw an error.

dlockhart: Ok cool 👍

dlockhart (approved):

PR #271, opened 2020-09-22
Override the body style to remove extra scrollbar

ianlaidlaw:
LMS pr: https://github.com/Brightspace/lms/pull/1597

See this pr for more info: https://github.com/Brightspace/brightspace-integration/pull/2496

image

### DIFF
--- a/d2l-sequence-viewer/d2l-sequence-viewer.js
+++ b/d2l-sequence-viewer/d2l-sequence-viewer.js
@@ -460,6 +460,9 @@ class D2LSequenceViewer extends mixinBehaviors([
 			return;
 		}
 
+		// Override the 'scroll' property set by the page to remove disabled scrollbar
+		document.body.style.overflow = 'hidden';
+
 		if (!entity) {
 			PerformanceHelper.perfMark('mark-api-call-start');
 		} else {


d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -460,6 +460,9 @@ class D2LSequenceViewer extends mixinBehaviors([
 			return;
 		}
 
+		// Override the 'scroll' property set by the page to remove disabled scrollbar
+		document.body.style.overflow = 'hidden';
```
imcdonald-d2l, line 464: Curious if setting it to `auto` would be better and still hide the scroll bar when not-needed? That way it will scroll if needed instead of hiding it. 🤔

ianlaidlaw: Hmm, possibly. I believe it _should_ never be present. I tested with auto and it didn't show up, but maybe it's safer to have that in case something does extend beyond the length of the screen somehow

imcdonald-d2l: Agreed, it _should_ never be present but there could be some weird case we are un-aware of. Could be useful to mitigate potential defects reported. Or we go with the approach of expecting no issues, leaving it set to `hidden` and fixing defects if there are any. I'm on the fence on what we should do... any opinions @pbrown-d2l ?

pbrown-d2l: I think this is totally fine this way. As long as it doesn't mess with Ian M's large HTML file fix where the scroll appears on the right. I don't think it will though

pbrown-d2l (approved):

d2l-sequence-viewer/d2l-sequence-viewer.js
imcdonald-d2l: This won't affect the fix I put up for HTML scrolling so we're good 👍

imcdonald-d2l (approved):

PR #272, opened 2020-10-08
DE39481 - Added mirrored styling for rtl

brissons:
This PR addresses [this defect](https://rally1.rallydev.com/#/detail/defect/397757644020) by adding styling for RTL languages. It is in conjunction with [this PR](https://github.com/BrightspaceUI/core/pull/890), which deals with the circular progress bar and the previous/next arrow icons.

This PR makes CSS changes to:

- Header element spacing and behavior ('Back to Content' moving, 'X', ellipsis)
- Mobile sidebar flyout direction
- Sidebar spacing (icons, arrows, completion count (old progress bar))
- End of unit page (icons and labels)

Down the road, it would be nice to use [logical properties ](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) so that we don't have to constantly change styling between left/right. However, we'll have to wait until it is more widely supported.

### DIFF
--- a/components/d2l-sequences-content-link-new-tab.js
+++ b/components/d2l-sequences-content-link-new-tab.js
@@ -11,6 +11,9 @@ export class D2LSequencesContentLinkNewTab extends D2L.Polymer.Mixins.Sequences.
 	static get template() {
 		return html`
 		<style>
+			:host {
+				--svg-center-offset: 60px;
+			}
 			.content-link-new-tab-container {
 				padding-top: 100px;
 				text-align: center;
@@ -27,12 +30,12 @@ export class D2LSequencesContentLinkNewTab extends D2L.Polymer.Mixins.Sequences.
 			}
 
 			#linkImage {
-				margin-left: 60px;
+				margin-left: var(--svg-center-offset);
 			}
 
 			:host(:dir(rtl)) #linkImage {
 				margin-left: initial;
-				margin-right: 60px;
+				margin-right: var(--svg-center-offset);
 				-webkit-transform: scale(-1, 1);
 				transform: scale(-1, 1);
 				transform-origin: center;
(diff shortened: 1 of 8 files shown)


components/d2l-sequences-content-link-new-tab.js
```diff
@@ -26,13 +26,21 @@ export class D2LSequencesContentLinkNewTab extends D2L.Polymer.Mixins.Sequences.
...
 
-			#linkImage{
-				margin-left:60px;
+			#linkImage {
+				margin-left: 60px;
+			}
+
+			:host(:dir(rtl)) #linkImage {
+				margin-left: initial;
+				margin-right: 60px;
```
ianlaidlaw, line 35: I think if we ever reuse a value like this we should put it into a CSS var. If we ever have to change 60px to 50px, we might miss updating the rtl version and then they could be out of sync.

brissons: Makes sense. There are a few cases where this would be useful.

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -93,6 +93,10 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 				padding-right: 5px;
 				width: 90%;
 			}
+			:host(:dir(rtl)) #title-container {
+				padding-right: initial;
+				padding-left: 5px;
```
ianlaidlaw, line 98: nit: looks like this 5px is repeated as well

ianlaidlaw (approved):

d2l-sequence-navigator/d2l-activity-link.js
brissons: Thanks!

PR #277, opened 2020-11-02
Added property for enabling media downloads

ktonon:
`<d2l-labs-media-player>` now has a property `allow-download` with obvious meaning. This PR adds a property to the sequence viewer which is forwarded to the media player.

See related PR https://github.com/Brightspace/lms/pull/3027

### DIFF
--- a/components/d2l-sequences-content-audio.js
+++ b/components/d2l-sequences-content-audio.js
@@ -15,7 +15,7 @@ export class D2LSequencesContentAudio extends D2L.Polymer.Mixins.Sequences.Autom
 			}
 		</style>
 		<template is="dom-if" if="[[useMediaPlayer]]">
-			<d2l-labs-media-player src="[[_fileLocation]]"></d2l-labs-media-player>
+			<d2l-labs-media-player src="[[_fileLocation]]" allow-download="[[allowMediaDownloads]]"></d2l-labs-media-player>
 		</template>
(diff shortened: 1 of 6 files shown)


ktonon:
@Jevan5 updated media-player to 1.16.0 in package lock file

Screen Shot 2020-11-05 at 2 26 53 PM

ktonon:
> Will this render as `<d2l-labs-media-player ... allow-download="false">` if `allowMediaDownloads` is false?

See https://github.com/Brightspace/lms/pull/3027, which omits the `allow-download` flag when it is not availablle

components/d2l-sequences-content-audio.js
```diff
@@ -46,6 +46,11 @@ export class D2LSequencesContentAudio extends D2L.Polymer.Mixins.Sequences.Autom
 				type: Boolean,
 				reflectToAttribute: true,
 				value: false
+			},
+			allowMediaDownloads: {
+				type: Boolean,
+				reflectToAttribute: true,
```
ktonon, line 52: I believe this causes the attribute to be omitted if the bound value is `false`. Anyhow, I've tested with and without. Without the flag at the top level, the download is not shown

ktonon: Screen Shot 2020-11-05 at 2 43 37 PM

Jevan5 (approved):

PR #278, opened 2020-11-03
DE39488: Adding html-block to module description to render MathML

brissons:
This PR addresses [this defect](https://rally1.rallydev.com/#/detail/defect/397788029136). User-defined math equations do not render in module descriptions for Chrome and Edge. This is because the language used to generate the equations (MathML) is not supported by chromium browsers.

This PR wraps the module descriptions with a new web-component ([html-block](https://github.com/BrightspaceUI/core/tree/master/components/html-block)) which displays html and integrates MathJax, a library used for rendering MathML.

[Before](https://user-images.githubusercontent.com/52549862/97949963-ea4b5600-1d5a-11eb-8a3a-c759c511742c.png) and [after](https://user-images.githubusercontent.com/52549862/97950048-35656900-1d5b-11eb-8cb6-8928fa0ed20e.png).

### DIFF
--- a/components/d2l-sequences-content-module.js
+++ b/components/d2l-sequences-content-module.js
@@ -1,4 +1,5 @@
 import './d2l-sequences-module-name.js';
+import '@brightspace-ui/core/components/html-block/html-block.js';
 import { html } from '@polymer/polymer/lib/utils/html-tag.js';
 import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
 import { PolymerElement } from '@polymer/polymer/polymer-element.js';
@@ -32,9 +33,10 @@ export class D2LSequencesContentModule extends mixinBehaviors([
 				<d2l-sequences-module-name href="[[href]]" token="[[token]]"></d2l-sequences-module-name>
 			</h1>
 
-			<p id="description">
-				[[description]]
-			</p>
+			<d2l-html-block>
+				<template id="description">
+				</template>
+			</d2l-html-block>
 		</div>
 `;
 	}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "d2l-sequences",
-  "version": "2.2.20",
+  "version": "2.2.22",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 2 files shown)


pbrown-d2l (approved):

components/d2l-sequences-content-module.js
```diff
@@ -32,9 +33,10 @@ export class D2LSequencesContentModule extends mixinBehaviors([
 				<d2l-sequences-module-name href="[[href]]" token="[[token]]"></d2l-sequences-module-name>
 			</h1>
 
-			<p id="description">
-				[[description]]
-			</p>
+			<d2l-html-block>
+				<template id="description">
```
ianlaidlaw, line 37: forgive my stupid question but how is this working? where did the description go? looking at the docs for this new component:
```
<d2l-html-block>
  <template>
    <!-- some html -->
  </template>
</d2l-html-block>
```
I would think you would need to do this:
```
<d2l-html-block>
  <template>
    <p id="description">
	  [[description]]
	</p>
  </template>
</d2l-html-block>
```

CodeBaboon: I was about to ask the same thing!

brissons: `[[description]]` wasn't hooked up to anything, so the old setup would run properly with just `<p id="description"></p>`. The old setup was just injecting html in the `<p>` using the `description` id as an identifier. So this change uses the same setup, but injects the html in the `template` instead.

If we were to do:
```
<d2l-html-block>
  <template>
    <p id="description"></p>
  </template>
</d2l-html-block>
```
... the id can't be found because it's wrapped in the `<template>` (which from my understanding, hides everything including ids), then can't be injected. So now the html is directly being injected in the template.

That's just from my understanding though, unless there's a better way?

ianlaidlaw: interesting, yeah i looked closer at the file and there isn't even a description property. my suggestion would be to create one and then use that instead of setting the inner html of the `#description` element... that should solve the issue of not finding it cause it's wrapped in a template and imo is the more proper way to handle this.

i just have concerns about removing the `p` tag and setting html of a template element.

ianlaidlaw: this is all assuming what i'm suggesting even works - perhaps the original dev took this approach for a reason.

brissons: I vaguely remember trying that out in the beginning, but the html was escaping... I don't think I tried it with this new component though. I'll give it a shot.

ianlaidlaw (approved):

components/d2l-sequences-content-module.js
brissons: Using properties, it looks like polymer escapes the html (I think to protect against XSS attacks), which might explain why it was done this way in the first place.

ianlaidlaw: yeah reading the docs that's what it looks like. i think this approach is fine in that case. i was just originally thrown off by the deletion of `[[description]]`. i'm assuming the original author tried to do this the traditional way and then realized the same thing you did but never deleted it.

PR #279, opened 2020-11-11
Adjust style of d2l-labs-media-player

ktonon:
Hiding overflow causes issues on mobile in landscape mode. Styles applied to d2l-video are not needed by labs media player.

Screen Shot 2020-11-11 at 2 55 40 PM

Removing the overflow hide allows the page to be scrolled.

### DIFF
--- a/components/d2l-sequences-content-audio.js
+++ b/components/d2l-sequences-content-audio.js
@@ -8,7 +8,7 @@ export class D2LSequencesContentAudio extends D2L.Polymer.Mixins.Sequences.Autom
 	static get template() {
 		return html`
 		<style>
-			d2l-audio, d2l-labs-media-player {
+			d2l-audio {
 				width: 100%;
 				height: calc(100% - 12px);
 				overflow: hidden;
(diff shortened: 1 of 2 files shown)


ktonon:
Tested on Desktop Chrome as well, video is the same width as before

rschick:
Can you check with the feature flag off so it uses d2l-video?

Should we get someone from sequences team to confirm they are ok with this?

Jevan5:
It seems that the video is still being cropped, but from the top this time. This is what the video is supposed to look like:
Screen Shot 2020-11-11 at 3 12 11 PM

Maybe more space has to be allocated for the container within the page?

rschick:
@ktonon does your screenshot show the page scrolled to the bottom, or is that what the page looks like on load?

ktonon (commented):
There is a conditional check, just further down. The style is rendered regardless, but is harmless if there is no `d2l-audio` element to apply to

ktonon:
> @ktonon does your screenshot show the page scrolled to the bottom, or is that what the page looks like on load?

Yes, it shows what it looks like after scrolling to the bottom. On load you can only see the top of the video

Jevan5:
Looks good to me. Are we going to wait for approval from sequences team?

rschick (approved):

ktonon:
Hold off merging. I'm testing this again, and something does not look right.

ktonon:
> Hold off merging. I'm testing this again, and something does not look right.

Ok, seems to be ok. Going to merge

PR #281, opened 2020-11-18
US121814 - LX - Show Due Dates on Modules/Topics in Sequence Nav (Left Nav)

prajapac:
US121814: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F449550794448

- This PR adds Due Dates to LX. 
- It also has CSS adjustments so it looks good when Availability Dates are also present.
- Further there are media queries to make due dates and availability dates fit on smaller/mobile screens, i.e. they get split on two lines (flex column), so due dates are on the first line (left aligned), and availability dates are on the second line (right aligned)
- With the exception of when 'dueDate' is read from entity.properties, it has been renamed to 'dueDateTime' everywhere else

LTR:
image

image

image

RTL:
image

image

image


ChrisKraljevicD2L (approved):

brissons:
lgtm! The only thing I would say is that it looks like a lot of the date stuff (availability, due date etc.) is repeated a bit. I wonder if down the road it might be worth while to consolidate it somewhere?

prajapac:
> lgtm! The only thing I would say is that it looks like a lot of the date stuff (availability, due date etc.) is repeated a bit. I wonder if down the road it might be worth while to consolidate it somewhere?

I did try and do that, specifically by getting the _getDueDateTimeString() to the utils.js file. However it didn't work. Perhaps I did it wrong. I'd be happy to go over the approach I tried with you and/or @chrisd2l if you like?

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -351,6 +364,22 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
...
 	}
 
+	_getDueDateTimeString(properties) {
+		if (!properties) {
+			return;
+		}
+
+		const dueDateTime  = properties.dueDate;
+
+		if (!dueDateTime) {
```
ianlaidlaw, line 374: up to you, but if you wanted to handle both exit early cases with 1 statement you could do:
```
const { dueDate } = (properties || {}); or const dueDateTime = (properties || {}).dueDate
if ( !dueDate) {
  return;
}}
```

ianlaidlaw: The first part of the first line I also decided to just destructure `dueDate` out of properties instead of creating a new, similarly named variable.

ChrisKraljevicD2L:
> 
> 
> > lgtm! The only thing I would say is that it looks like a lot of the date stuff (availability, due date etc.) is repeated a bit. I wonder if down the road it might be worth while to consolidate it somewhere?
> 
> I did try and do that, specifically by getting the _getDueDateTimeString() to the utils.js file. However it didn't work. Perhaps I did it wrong. I'd be happy to go over the approach I tried with you and/or @ChrisKraljevicD2L if you like?

I'm down to take a look at the approach ya tried :+1:

d2l-sequence-navigator/d2l-activity-link.js
prajapac: That seems good, I can do that. However `dueDateTime` is intentional, as everywhere else I called it that. Too bad entity returns `dueDate`. The former is more semantically correct.

ianlaidlaw: IMO if the backend/api is calling it a certain thing, we should remain consistent with it. If we ever want to rename it to something more appropriate, then we should handle that at the source. Even though this is a very minor example, it does create a disconnect and some confusion

prajapac: Hmmm, it would require quite a bit of changes from `dueDateTime` to `dueDate` in this PR. The only place it is `dueDate` vs. `dueDateTime` is when the property is read from `properties`, i.e. `properties.dueDate`

brissons (approved):

imcdonald-d2l (approved):

PR #282, opened 2020-11-20
Adding tabindex and aria role to avail dates

ChrisKraljevicD2L:


d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -366,5 +366,9 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 		const { startDate, endDate } = properties;
 		return formatAvailabilityDateString(this.localize, startDate, endDate, true);
 	}
+
+	_getTabIndex(showDates) {
+		return showDates ? '0' : '-1';
+	}
```
ChrisKraljevicD2L, line 372: I don't want users to be able to tab to an empty div and open an empty tooltip when there's no dates.

prajapac (approved):

PR #290, opened 2020-12-01
Switch to GitHub Actions

dlockhart:
Second attempt at this one, hopefully now the incremental-release portion will work correctly!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
+      - 'v.*'
+jobs:
+  release:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    name: Release
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: Brightspace/third-party-actions@actions/checkout
+      - name: Setup Node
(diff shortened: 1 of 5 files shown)


imcdonald-d2l (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
+name: Release
+on:
+  push:
+    branches:
+      - master
+      - 'v.*'
```
svanherk, line 6: Is this to run the release workflow again after a tag is made?

svanherk: Won't that just skip if it was a release made by `incremental-release`, and cause _another_ release if someone makes a manual release for cert?

dlockhart: It was trying to match what they previously had in their `.travis.yml` file... which I'm now noticing didn't get deleted like it should have when I cherry-picked these commits. I'll remove it in a follow-up PR. Not sure though what should happen here with the `v.*` branches... I'm assuming they were being used for maintenance branches.

dlockhart: I don't actually see any `v.*` branches -- @imcdonald-d2l what's the typical hotfix workflow? Looks like it might be `releases/*`?

imcdonald-d2l: I'm not sure what our hotfix workflow is actually. @sarahboyd or @pbrown-d2l might know more information on this.

PR #291, opened 2020-12-01
GHA cleanup

dlockhart:
Removes the `.travis.yml` file which got missed in my cherry-pick in the last PR.

Also -- and here's where I need your expertise @imcdonald-d2l -- this switches the release workflow from running on `master` and `v.*` branches to `master` and `release/*`. This seems to better match what you've been doing in practice with your hotfix branches, but let me know if not.

Lastly, I might recommend switching the default increment from `patch` to `minor`. Otherwise for your release branches you may not be able to slide in a patch release in between since the next patch release may have already occurred by the time you realize you need to hotfix something. Let me know and I can switch that as well.


imcdonald-d2l (commented):
Looks good to me but would like @sarahboyd or @pbrown-d2l to comment on what they think on the release and hotfix stuff you mentioned. I'm fine switching from `patch` to `minor` for the sake of improved hotfix-ability

sarahboyd (approved):
Switching from  `patch` to `minor` sounds like a good plan to me. Thanks for doing this Dave!

imcdonald-d2l (approved):

sarahboyd (approved):

dlockhart:
Alright, I've switch it so that `minor` is the default. Thanks all!

PR #295, opened 2020-12-07
DE41622 aria-label and title on availability dates

ChrisKraljevicD2L:
[DE41622](https://rally1.rallydev.com/#/detail/defect/462212718384?fdp=true): LX > A11Y labels for availability dates


d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -384,7 +402,15 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 			return;
 		}
 		const { startDate, endDate } = properties;
-		return formatAvailabilityDateString(this.localize, startDate, endDate, true);
+		return formatAvailabilityDateString(this.localize, startDate, endDate, availDateTooltipSuffix);
+	}
+
+	_getAvailabilityDateAriaLabel(properties) {
+		if (!properties) {
+			return;
```
ianlaidlaw, line 410: question, should these be returning `''`? `_getAvailabilityDateAriaLabel` computes the string value for something, would anything weird happen if the function just returned nothing?

ianlaidlaw (approved):

d2l-sequence-navigator/d2l-activity-link.js
ChrisKraljevicD2L: That's fair, I can switch these to return empty strings :+1: for both sighted and non-sighted users I think the end result would be the same BUT if we ever write additional code that uses this stuff, returning strings would be good for consistency

imcdonald-d2l (approved):

brissons (approved):

PR #276, opened 2020-10-30
US121812 LX Availability Dates

ChrisKraljevicD2L:
[US121812](https://rally1.rallydev.com/#/289692574792/dashboard?detail=%2Fuserstory%2F449549534976&fdp=true)

image

image

PR #300, opened 2020-12-17
Availability date tool tip behaviour not consistent on iOS

ianlaidlaw:
There were a few problems with the tooltip in LX that this pr will fix:

1. On mobile, tapping on the avail dates to make the tooltip show was difficult as the tap area was very small.
2. Even if you tapped directly in the middle, the tap would go behind the date and perform the action of the element behind it. This could be expanding/collapsing, or even navigating into LX.
3. Tabbing to the avail dates (not mobile) was broken.

### DIFF
--- a/d2l-sequence-navigator/d2l-activity-link.js
+++ b/d2l-sequence-navigator/d2l-activity-link.js
@@ -233,6 +233,7 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 							role="note"
 							aria-label$="[[_availabilityDateAriaLabel]]"
 							title$="[[_availabilityDateAriaLabel]]"
+							onClick="[[_onDatesClick]]"
 						>
 							[[_availabilityDateString]]
 						</div>
@@ -387,6 +388,12 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 			return;
 		}
 		return getDueDateTimeString(properties.dueDate, this.localize);
+	}
+
+	// Prevent navigating/collapsing the background element when
+	// attempting to touch the dates for viewing the tooltip.
+	_onDatesClick(e) {
+		e.stopPropagation();
 	}
 
 	_getAvailabilityDateString(properties) {
--- a/d2l-sequence-navigator/d2l-activity-link.js
+++ b/d2l-sequence-navigator/d2l-activity-link.js
@@ -168,7 +168,6 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 			#due-date-time, #availability-dates {
 				color: var(--d2l-color-ferrite);
 				font-size: 0.65rem;
-				line-height: var(--d2l-body-small-text_-_line-height);
 			}
 
 			@media (max-width: 525px) {
--- a/d2l-sequence-navigator/d2l-activity-link.js
+++ b/d2l-sequence-navigator/d2l-activity-link.js
@@ -240,7 +240,6 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 						<d2l-tooltip
 							for="availability-dates"
 							boundary="[[_availDateTooltipBoundary]]"
-							close-on-click="true"
 						>
 							[[_availabilityDateTooltip]]
 						</d2l-tooltip>
--- a/d2l-sequence-navigator/d2l-activity-link.js
+++ b/d2l-sequence-navigator/d2l-activity-link.js
@@ -240,9 +240,7 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 						<d2l-tooltip
 							for="availability-dates"
 							boundary="[[_availDateTooltipBoundary]]"
-						>
-							[[_availabilityDateTooltip]]
-						</d2l-tooltip>
+						>[[_availabilityDateTooltip]]</d2l-tooltip>
 					</div>
 				</div>
 			</template>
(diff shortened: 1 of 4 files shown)


d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -168,7 +168,6 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 			#due-date-time, #availability-dates {
 				color: var(--d2l-color-ferrite);
 				font-size: 0.65rem;
-				line-height: var(--d2l-body-small-text_-_line-height);
```
ianlaidlaw, line 171: deleting this makes the height a little bigger, making tapping on the dates easier

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -317,7 +319,12 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 				type: String,
 				value: '',
 				computed: '_getAvailabilityDateAriaLabel(entity.properties)'
-			}
+			},
+			_showDates: {
```
ianlaidlaw, line 323: I think when making the tooltip, this function was missed in some files. It was attempting to call this for tabIndex, but this property/function did not exist

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -390,6 +397,22 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 		return getDueDateTimeString(properties.dueDate, this.localize);
 	}
 
+	// Prevent navigating/collapsing the background element when
+	// attempting to touch the dates for viewing the tooltip.
+	_onDatesClick(e) {
+		e.stopPropagation();
+	}
+
+	_getShowDates(properties) {
```
ianlaidlaw, line 406: I got this from sequence launcher fra: https://github.com/Brightspace/sequence-launcher-fra/blob/master/components/d2l-sequence-launcher-accordion.js#L344

ianlaidlaw: One difference is I shouldn't check for due dates. If there was a due date but not avail dates, this would have returned true and shown an empty tooltip when there wasn't actually an avail date

d2l-sequence-navigator/d2l-outer-module.js
```diff
@@ -675,6 +682,22 @@ class D2LOuterModule extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 		return getDueDateTimeString(properties.dueDate, this.localize);
 	}
 
+	// Prevent navigating/collapsing the background element when
```
ianlaidlaw, line 685: unfortunately there's lots of duplicated code with all the different files 😞

d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -390,6 +397,22 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 		return getDueDateTimeString(properties.dueDate, this.localize);
 	}
 
+	// Prevent navigating/collapsing the background element when
+	// attempting to touch the dates for viewing the tooltip.
+	_onDatesClick(e) {
+		e.stopPropagation();
```
ianlaidlaw, line 403: this is what prevents the click event from propagating up and causing the click to be registered by the element behind the availability dates text

d2l-sequence-navigator/d2l-lesson-header.js
```diff
@@ -447,9 +449,9 @@ class D2LLessonHeader extends ASVFocusWithinMixin(CompletionStatusMixin()) {
 			return false;
 		}
 
-		const { startDate, endDate, dueDate } = properties;
+		const { startDate, endDate } = properties;
 
-		return startDate || endDate || dueDate;
+		return startDate || endDate;
```
ChrisKraljevicD2L, line 454: Switching this to only check availability dates will break CSS for lesson-header, since [`_getDateContainerClasses`](https://github.com/BrightspaceHypermediaComponents/sequences/blob/ilaidlaw/ios-tooltip/d2l-sequence-navigator/d2l-lesson-header.js#L246) takes in `_showDates`. If there is only a due date but no availability dates, the wrong classes will be applied.

I suggest altering `_getTabIndex` to take in `entity.properties` so it can work on only avail dates while `_showDates` can still use `dueDate`.

ChrisKraljevicD2L (changes requested):

d2l-sequence-navigator/d2l-lesson-header.js
ChrisKraljevicD2L: Actually, your solution in [`sequence-launcher-fra`](https://github.com/Brightspace/sequence-launcher-fra/pull/178/files#r545283737) would work here too :+1:

d2l-sequence-navigator/d2l-activity-link.js
brissons: It looks like this was [added with availability dates](https://github.com/BrightspaceHypermediaComponents/sequences/pull/276). Double checking that removing this won't cause any ui formatting issues?

d2l-sequence-navigator/d2l-lesson-header.js
ianlaidlaw: good catch, ill adjust accordingly

d2l-sequence-navigator/d2l-activity-link.js
ianlaidlaw: yeah, it does slightly increase the height of the text area a little bit. i think i'm going to revert this because i don't think sarah realized that tap events were going through the text, I think maybe she thought the text area was really small and it was difficult to press it without accidentally pressing above or below it. so since i blocked those touch events from going behind, this isn't really doing much

ianlaidlaw: increasing the height was a suggestion from Sarah + Tara. but I don't think we'll need to mess with the designs to fix this

ChrisKraljevicD2L (approved):

PR #312, opened 2021-06-18
DE43897 - fixing Back to Content button completion tracking

quinnbudan:
## Relevant Rally Links

- [DE43897](https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fdefect%2F601409434557&fdp=true?fdp=true): Lessons/New Learner experience/Content Statistics > 20.21.05 > Total Time Spent and Average Time Spent not calculating for users content topic visits

## Description

When the "Back to Content" button was clicked, the automatic completion tracking mixin wouldn't work due to the timeout around the completion tracking action being called. This timeout is necessary to allow for completion queues to clear.

## Goal/Solution

Here I'm reworking the automatic completion tracking mixin to allow for us to specify the timeout. Then, when the "Back to Content" button is clicked, a `back-to-content` event is fired and handled by immediately calling the completion tracking action.

## Video/Screenshots

Statistics before fix:
before_back_to_content_fix

Statistics after fix:
after_back_to_content_fix

## Related PRs

N/A

## Remaining Work
- [ ] Fix completion tracking on back button in browser pressed before navigating around LX, see [DE44070](https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fdefect%2F602131736115&fdp=true?fdp=true): Lessons/LX/Content Statistics > Total Time Spent and Average Time Spent not calculating for users content topic visits when browser back button pressed

### DIFF
--- a/mixins/d2l-sequences-completion-tracking-mixin.js
+++ b/mixins/d2l-sequences-completion-tracking-mixin.js
@@ -28,15 +28,21 @@ function CompletionTrackingMixin() {
 			};
 		}
 
+		getFinishCompletionHandler({ delay = FINISH_COMPLETION_DELAY } = {}) {
+			// we don't want any chance of a delay if delay was set to 0,
+			// see https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout
+			return delay > 0 ?
+				() => setTimeout(() => this.finishCompletion(), delay) :
+				() => this.finishCompletion();
+		}
+
 		finishCompletion() {
 			if (!this._completionEntity || this._skipCompletion) {
 				return;
 			}
 
-			setTimeout(() => {
-				this._performViewActions(this._completionEntity, 'finish-view-activity')
-					.then(() => { this._completionEntity = null; });
-			}, FINISH_COMPLETION_DELAY);
+			this._performViewActions(this._completionEntity, 'finish-view-activity')
+				.then(() => { this._completionEntity = null; });
 		}
 
 		startCompletion() {
--- a/d2l-sequence-viewer/d2l-sequence-viewer.js
+++ b/d2l-sequence-viewer/d2l-sequence-viewer.js
@@ -580,6 +580,7 @@ class D2LSequenceViewer extends mixinBehaviors([
 
 	_onClickBack() {
 		this.telemetryClient.logTelemetryEvent('back-to-content');
+		this.dispatchEvent(new CustomEvent('back-to-content', { bubbles: true, composed: true }));
 
 		if (!this.backToContentLink) {
 			return;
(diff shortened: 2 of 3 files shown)


brissons (approved):
lgtm! Tested and it looks like pressing the 'Back to Content' button is now allowing the average time to be calculated property.

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -28,15 +28,21 @@ function CompletionTrackingMixin() {
 			};
 		}
 
+		getFinishCompletionHandler({ delay = FINISH_COMPLETION_DELAY } = {}) {
```
ianlaidlaw, line 31: any reason you wrapped this up in an object? would this not have worked otherwise?
i think you could do the same thing by doing:
```
getFinishCompletionHandler(delay = FINISH_COMPLETION_DELAY) {
```
and
```
this._immediateFinishCompletionCallback = this.getFinishCompletionHandler(0).bind(this);
```

ianlaidlaw: I can see the benefit by having a named variable in that other spot: `this._immediateFinishCompletionCallback = this.getFinishCompletionHandler({ **delay**: 0 }).bind(this);` so it's clear what the 0 means, was that the intention?

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -28,15 +28,21 @@ function CompletionTrackingMixin() {
...
+			// we don't want any chance of a delay if delay was set to 0,
+			// see https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout
```
ianlaidlaw, line 33: I'm assuming this comment is referring to the `Late timeouts` section. super weird that MDN doesn't have linking to specific sections

ianlaidlaw (approved):
L G T M !!!

mixins/d2l-sequences-completion-tracking-mixin.js
quinnbudan: that was exactly the intention! I'm happy to not wrap it though

mixins/d2l-sequences-completion-tracking-mixin.js
quinnbudan: That was the section, if I make another change I can specify the section in the comment cause it is pretty hard to find

mixins/d2l-sequences-completion-tracking-mixin.js
ianlaidlaw: nah, i like the idea of having the named variable!

PR #316, opened 2021-07-23
DE44141: Apply styling when router clears content

brissons:
## Relevant Rally Links

- [DE44141](https://rally1.rallydev.com/#/289692574792/custom/355050439968?detail=%2Fdefect%2F602447944189&fdp=true): LX > Sequence Viewer > Sudden enlarged page when click back navigation

## Description

If you're on the 'end of unit' page and navigate back to an html topic, the frame will suddenly enlarge before loading the html topic.

This is because when the entity changed, `this.isNotHtmlEntity` was calculated and toggled html topic specific `css` styles while the previous view was still loaded. This was causing the html topic view's `css` (or more accurately, lack of `css`) to be shown briefly in the end of unit page before loading the html topic, hence the enlargement. This happens when transitioning from other views as well, but it's less obvious.

I added an event which fires when the content view is removed before the next view is loaded. This triggers the `this.isNotHtmlEntity` calculation while there is no content rendered, avoiding the `css` jittering.

## Goal/Solution

Have the required css styling applied before the view is loaded to avoid sudden styling changes.

## Video/Screenshots

Before:
before

After:
after

## Remaining Work


ianlaidlaw (approved):
LGTM! i will dev test!

prajapac (approved):
Dev Tested, LGTM!

PR #317, opened 2021-07-23
Use regexp to fetch all subentities for module children

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F605009240383

## Description

[A change in node-siren-parser](https://github.com/Brightspace/node-siren-parser/commit/11a97f2ff7832eabac66dfb8d44f150e45fff6fd#diff-9c0ec1b0a889e599f3ff81590864dd9dc65684a86b41f1da75acc53fafed12e3R30) no longer allowed us to pass `undefined` as a rel for fetching subEntities. We did this as they could be modules or topics. This changes that to a regExp to accept all types (the same as `undefined`, basically) and then filters them appropriately. I also extracted all instances of this into a helper as they were all more or less the same.

image


helpers/hypermedia-helper.js
```diff
@@ -115,6 +115,23 @@ class HypermediaHelper {
...
+		return entity.getSubEntities(new RegExp('[\\s\\S]*'))
+			.filter(entity =>
+				(entity.href &&
+					(
+						entity.hasClass('sequenced-activity') &&
+						entity.hasClass('available') &&
+						!entity.hasClass('unavailable')
+					) ||
+					(entity.hasClass('sequence-description'))
+				));
```
ianlaidlaw, line 133: originally this is where the `getHref` function would have been applied to every entity but like... If you look at it, it didn't actually do anything:
```
return entity && entity.getLinkByRel && entity.getLinkByRel('self') || entity || '';
```
We were never wanting to use the `href`, ever. We always wanted the whole entity. So the `|| entity` got executed every time, essentially doing nothing.

I think this was a hold over when we rewrote how this behaved last year re: lazy loading.

helpers/hypermedia-helper.js
```diff
@@ -115,6 +115,23 @@ class HypermediaHelper {
...
 		}
 	}
+
+	static getTopicAndModuleSubEntities(entity) {
+		if (!entity) {
+			return [];
+		}
+
+		return entity.getSubEntities(new RegExp('[\\s\\S]*'))
+			.filter(entity =>
```
ianlaidlaw, line 125: All the `.filter` functions in each file were slightly different and I'm not sure why. We basically want them all to do the same thing, and I think what I have here captures that.

helpers/hypermedia-helper.js
ianlaidlaw: `entity.getLinkByRel` always returned `false`

helpers/hypermedia-helper.js
ianlaidlaw: 1. does the entity have an href? kind of a pointless check, of course it does. if it doesnt, them something has gone horribly wrong in the serializer
2. is it a sequenced-activity? if it is, it must be marked available
3. is it a sequence-description (module)?

helpers/hypermedia-helper.js
```diff
@@ -115,6 +115,23 @@ class HypermediaHelper {
 			return [];
 		}
 	}
+
+	static getTopicAndModuleSubEntities(entity) {
+		if (!entity) {
+			return [];
+		}
+
+		return entity.getSubEntities(new RegExp('[\\s\\S]*'))
```
ianlaidlaw, line 124: this regexp is basically "everything"

test/d2l-sequence-navigator/data/unit3.json
```diff
@@ -10,7 +10,7 @@
 			"href": "data/unit3-activity1.json"
 		},
 		{
-			"class": ["sequence", "sequenced-activity"],
+			"class": ["sequence", "sequenced-activity", "available"],
```
ianlaidlaw, line 13: This was failing a test since I strengthened the check. For units specifically, it was checking only if it's children did not have the `"unavailable"` class, while some other levels were checking if it did have the `"available"` class.

All the test data not including this is out of date as we can be confident it will have one or the other. So available topics will have this class. But I'm not going to worry about updating _all_ the test data right now.

image

ianlaidlaw: maybe my check doesn't need to check both values, but it's fine

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 on: pull_request
 jobs:
   test:
-    timeout-minutes: 15
+    timeout-minutes: 20
```
ianlaidlaw, line 5: I bumped this as frequently I was getting builds timing out which would have finished around 16 minutes or so. This definitely isn't ideal to do, but I think a good Dragon Slayer idea would be to investigate why the Unit Tests step takes so long.

test/d2l-sequence-navigator/d2l-sequence-navigator.html
```diff
@@ -26,7 +26,7 @@
 		<script type="module">
 import '@polymer/polymer/polymer-legacy.js';
 import '../../d2l-sequence-navigator/d2l-sequence-navigator.js';
-import SirenFixture from 'polymer-siren-test-helpers';
+import SirenFixture from '/polymer-siren-test-helpers.js';
```
ianlaidlaw, line 29: needed to run tests locally

coffeexcode (approved):
LGTM. I like the refactoring into HypermediaHelper

brissons (approved):
Code LGTM! Will dev test now.

prajapac (approved):
LGTM, good to merge pending @brissons' Dev Testing!

brissons:
LGTM! I double checked the order and visibility of a bunch of topics and it looks accurate.

PR #326, opened 2021-08-11
DE44003 - Better error handling on failed completion tracking

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F601854468893&fdp=true

## Description

When changing tabs/windows in LX while viewing a topic that has completion time tracking, "start completion" is called on that topic. However, the entity very unreliably has the correct action associated with that. It will either have all of: `view-activity`, `view-activity-duration`, `set-last-viewed-content-object` or just: `finish-view-activity` depending on what state it is in. Check `SirenContentSerializer.cs` for that logic.

As a result, this will result in an uncaught promise about 50-75% of the time in my experience depending on the speed or whatever the user changes focus. If the action cannot be found it will reject the promise, and then this rejection goes unhandled by the caller. We used to do logic in a catch, but it was removed. I added a no-op catch to simply swallow the rejected promise. There could affect our logging in a minor way as errors that happen during the `startCompletion` flow will not be logged, but I feel okay doing this as there used to be a catch here anyways. The true fix here is to investigate what is going on with the entity actions.

image

## Video/Screenshots

Before, this is the user changing tabs back and forth with LX.

completionerrors


ChrisKraljevicD2L (approved):

mixins/d2l-sequences-completion-tracking-mixin.js
```diff
@@ -63,7 +64,7 @@ function CompletionTrackingMixin() {
 					);
 
 				if (action.isNothing()) {
-					return reject(entity, 'no action found');
+					return reject(new Error('no action found'));
```
ianlaidlaw, line 67: I haven't been able to recreate it as maybe the console is different than our logging software, but I _think_ this is where that `[object Object]` error comes from. That happens when you attempt to stringify an object improperly with a string.
https://stackoverflow.com/questions/10156309/error-object-object-on-javascript
Also I updated it to return an Error as that is what MDN docs suggest:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/reject
> The static Promise.reject function returns a Promise that is rejected. For debugging purposes and selective error catching, it is useful to make reason an instanceof Error.

coffeexcode (approved):

brissons (approved):

prajapac (approved):

PR #328, opened 2021-08-19
US130912: LX > Investigate extremely long Build times

coffeexcode:
## Relevant Rally Links

- [US130912](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F606827524297&fdp=true?fdp=true): LX > Investigate extremely long Build times

## Description
This PR is to reduce the time it takes for CI to run the test step. Based on the Brightspace support browsers found [here](https://documentation.brightspace.com/EN/brightspace/requirements/all/browser_support.htm) and the related native support for shadow DOM elements [here](https://caniuse.com/shadowdomv1) we should be able to safely reduce the numbers of tests for this repository by only testing the components using the browers native shadow DOM. 

Before, we would be testing the same component multiple times using the native shadow DOM, along with a polyfill, and again using `lazyRegister` which doesn't seem to be supported in Polymer 3. Some components we're also being tested a 4th time with no explicit properties (defaulting to native shadow DOM if available, otherwise polyfilled)

## Results
Based on the recent CI runs, and the latest CI run from this PR, we should see a reduction of CI times by up to 80% (>50% reduction for all runs)

Before:
image

After:
image


ianlaidlaw:
look at those savings!!

test/testSuites.js
```diff
@@ -0,0 +1,80 @@
+export const allTests = [
```
ianlaidlaw, line 1: jw are these being used anywhere?

ianlaidlaw (approved):

test/testSuites.js
coffeexcode: Not really, I did think about removing it, but if someone wanted to test locally using the full suite then they'd have to pull in all the test cases from somewhere. CI now only uses the `nativeShadowDomTests`

ianlaidlaw: fair enough, no harm keeping it like this in case people want to use them

PR #332, opened 2021-09-07
Update telemetry to record absolute URL

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=/userstory/608174014131

## Description

> in the 20.21.10 release Telemetry events will have the URL enforced. In LX we pass the ID in the URL field. That needs to change to an absolute URL in order to conform to the event requirements.
> 
> 
> Here is the file that likely needs to be upgrades:
> 
> https://search.d2l.dev/xref/BrightspaceHypermediaComponents/sequences/helpers/telemetry-helper.js?r=85091531&fi=logTelemetryEvent#logTelemetryEvent
> 
> 
> When the below method is called, the first parameter should be an absolute URL the user is currently viewing.
> 
> `.setObject(encodeURIComponent(id), 'Sequence Viewer', id);`
>
> This improvement should be done for both Telemetry and Performance events.

## Goal/Solution

In `eventBody.js`, the signature for `setObject` is:

```javascript
setObject(id, type, url, value) {
```

In the telemetry and performance logging events, the URL parameter is changed to output `window.location.href`, which is the currently viewed URL (see screenshot).
The id parameter will be passed the id and the type is unchanged.

## Video/Screenshots

Temporary alert showing what would be recorded in the URL field of the event:

Telemetry Test

## Related PRs

- N/A

## Remaining Work


helpers/telemetry-helper.js
```diff
@@ -21,7 +21,7 @@ class TelemetryHelper {
 
 		const eventBody = new d2lTelemetryBrowserClient.EventBody()
 			.setAction('Created')
-			.setObject(encodeURIComponent(id), 'Sequence Viewer', id);
+			.setObject(id, 'Sequence Viewer', window.location.href);
```
pbrown-d2l, line 24: based on the comment [here](https://search.d2l.dev/xref/Brightspace/d2l-telemetry-browser-client/) it looks like the id might still need to be encoded to avoid certain characters due to a limitation in the BDP processing

pbrown-d2l (changes requested):

coffeexcode (approved):

ianlaidlaw (approved):

brissons (approved):

pbrown-d2l (approved):

PR #339, opened 2021-11-02
US131564 sequences back button text

kristian-torres:
## Relevant Rally Links

[US131564](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F608870749955&fdp=true) - Update language of LX 'Back' button UI changes

## Description

We want to support the following 2 requirements in regard to the Back button in LX:
1. When the user will be returned to the normal 'Content' (sequence launcher or classic content) then the button reads "Back to Content".
2. When the user will return elsewhere such as an external system or the homepage, then the buttons read "Back".

For testing, I have added a widget and a link on top of the normal sequence launcher path to get into LX. In all scenerios except going through the sequence launcher, we will display the 'Back' text on both the navigation bar as well as (when necessary) the end of sequences button at the End of Unit page.

## Goal/Solution

Look for the added class in the `default-return-url` and `end-of-sequences` siren link and create appropriate templates (and supporting template functions/properties) to logically show either `Back To Content` or a newly added localized `Back` text.

## Video/Screenshots

US131564-Back
US131564-BackToContent

## Related PRs

[Kristian torres/us131564 serializer update for default user return location](https://github.com/Brightspace/lms/pull/15697)


components/d2l-sequences-content-eol-main.js
```diff
@@ -93,10 +93,18 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
...
 
 				<template is="dom-if" if="[[showEnhancedView]]">
-					<d2l-missed-activity href="{{href}}" token="[[token]]"> </d2l-missed-activity>
-					<d2l-button on-click="_onClickBack">
-						[[localize('backToContents')]]
-					</d2l-button>
+					<template is="dom-if" if="[[showDefaultReturnButtonText]]">
+						<d2l-missed-activity href="{{href}}" token="[[token]]"> </d2l-missed-activity>
+						<d2l-button on-click="_onClickBack">
+							[[localize('backToContents')]]
```
prajapac, line 99: I knew this at one point, but do you know what the difference between 'backToContent' and 'backToContents' is?

prajapac (approved):

components/d2l-sequences-content-eol-main.js
```diff
@@ -93,10 +93,18 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 				</template>
 
 				<template is="dom-if" if="[[showEnhancedView]]">
-					<d2l-missed-activity href="{{href}}" token="[[token]]"> </d2l-missed-activity>
-					<d2l-button on-click="_onClickBack">
-						[[localize('backToContents')]]
-					</d2l-button>
+					<template is="dom-if" if="[[showDefaultReturnButtonText]]">
```
ianlaidlaw, line 96: if im not mistaken, the only difference is the button text, so i think we can move this template to a more specific location so we're not repeating a bunch of stuff:
```
<d2l-button on-click="_onClickBack">
	<template is="dom-if" if="[[showDefaultReturnButtonText]]">
		  [[localize('backToContents')]]
    </template>
    <template is="dom-if" if="[[!showDefaultReturnButtonText]]">
		  [[localize('back')]]
    </template>
</d2l-button>
						
```

components/d2l-sequences-content-eol-main.js
```diff
@@ -220,6 +233,14 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 		this._poller.teardownPolling();
 	}
 
+	_isDefaultUserReturnLocation(entity) {
+		const defaultReturnUrl = entity && entity.getLinkByRel('https://sequences.api.brightspace.com/rels/end-of-sequence') || '';
```
ianlaidlaw, line 237: `defaultReturnUrl` looks to be a SirenLink, which is an object so I think defaulting it to an empty string if the `getLinkByRel` fails wouldn't be appropriate. I think it would be better to just move one of the checks lower. Also the name doesn't _exactly_ match what it is yet, as it isn't a url, it's a SirenLink for End of Sequence.
```
const endOfSequenceLink = entity && entity.getLinkByRel( ... );

if (endOfSequenceLink && endOfSequenceLink.class.includes('default-user-return-location') { .... }
```
You can check to see if `class` exists too if you want but IMO isn't super necessary, we should be confident it exists

components/d2l-sequences-content-eol-main.js
```diff
@@ -220,6 +233,14 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
...
+		if (defaultReturnUrl.class && defaultReturnUrl.class.includes('default-user-return-location')) {
+			return true;
+		}
+		return false;
```
ianlaidlaw, line 241: nit: could be a ternary if you wanted

lang/en.json
```diff
@@ -1,6 +1,7 @@
 {
     "activitiesFinishedGreatJob": "You have completed all of the activities in this unit. Great job!",
     "activity": "Activity",
+	"back": "Back",
```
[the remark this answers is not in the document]
kristian-torres: I think you're right. Maybe my tab spacing doesn't match the spacing when the code gets pushed? I can fix this.

components/d2l-sequences-content-eol-main.js
kristian-torres: not at all 🙃

components/d2l-sequences-content-eol-main.js
kristian-torres: that makes sense. I can make that change. I guess I was overly worried about removing certain properties, etc that I just duplicated the whole template. Thanks!

components/d2l-sequences-content-eol-main.js
ianlaidlaw: I see where the naming may have confused you, it could have come from the `_getBackToContentLink(entity)` which you edited but was setup similarly in the past. it's calling SirenLinks urls and defaulting to empty strings. That function is also wrong

d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -725,7 +732,10 @@ class D2LSequenceViewer extends mixinBehaviors([
 
 	_setBackButtonText() {
 		if (window.innerWidth > 435) {
-			this._navBackText = this.localize('backToContent');
+			this._navBackText = this.localize('back');
```
ianlaidlaw, line 735: small nit, but should this be in an `else` so we don't set this when it will just be overridden on line 737?

components/d2l-sequences-content-eol-main.js
kristian-torres: yeah, I was struggling to come up with a good name but what you mentioned clarifies exactly what I was trying to get.

components/d2l-sequences-content-eol-main.js
prajapac: If I'm remembering correctly, one is the button on the top left and one at the end of unit page.

components/d2l-sequences-content-eol-main.js
ianlaidlaw: i _think_ you could also simplify it further by doing the check right in this guy:
```
[[localize('back')]]
```
where it can check a property and render something different but off the top of my head i'm not sure what that would look like, or if it's even possible

components/d2l-sequences-content-eol-main.js
prajapac: I like that!

kristian-torres: me too. looks kinda awkward.

d2l-sequence-viewer/d2l-sequence-viewer.js
kristian-torres: oh i see. Yeah, normally I would do a default action which, in this case, would be to set it to back and only then override this if needed. But I can swing either way if multiple people prefer it :)

components/d2l-sequences-content-eol-main.js
kristian-torres: interesting. I'd love to change that but it would be a headache to get all those translated resource files changed as well...

components/d2l-sequences-content-eol-main.js
kristian-torres: @ChrisKraljevicD2L recommended a suggestion here where we computed the localized text and then pass that into a property that gets rendered within the d2l back button. However, I was hitting issues executing the localize function when the logic was being done, so I changed the approach back to what Ian suggested.

components/d2l-sequences-content-eol-main.js
```diff
@@ -149,6 +154,11 @@ export class D2LSequencesContentEoLMain extends D2L.Polymer.Mixins.Sequences.Ret
 			_pollInterval: {
 				type: Number,
 				value: 1000
+			},
+			_isDefaultReturnButton: {
```
ChrisKraljevicD2L, line 158: nit: `d2l-sequence-viewer.js`'s equivalent variable's name is `_hasDefaultReturnLocation`. Should probably get these two named the same.

ChrisKraljevicD2L (approved):

ianlaidlaw (approved):

brissons (commented):
Looks great! I'm wondering about the spacing between the back button and the separator (the vertical dots). Nothing I would block the pr for, but I'm wondering if it should be dynamic (ie. be closer or further depending on the text).

brissons:
> Looks great! I'm wondering about the spacing between the back button and the separator (the vertical dots). Nothing I would block the pr for, but I'm wondering if it should be dynamic (ie. be closer or further depending on the text).

If so, this will likely be a separate ticket, so please ignore :)

components/d2l-sequences-content-eol-main.js
kristian-torres: Changed now!

brissons (approved):

PR #340, opened 2021-11-11
Spacing for no default return url on back button

coffeexcode:
## Relevant Rally Links

- [US133556](https://rally1.rallydev.com/#/?detail=/userstory/614158191877&fdp=true): Update spacing around the Back button after the language change

## Description
When the back button has a non-default return location (i.e. Does not go back to content) the short text is aligned poorly with the separator and flyout menu. When the button only says back, we needed to change the spacing by overriding some of the default margins with specific margins.

## Screenshots
Before:
image

After:
image


d2l-sequence-viewer/d2l-sequence-viewer-header.js
```diff
@@ -139,16 +145,16 @@ PolymerElement) {
 			<div id="container">
 				<div id="header-left">
 					<div id="header-left-inner">
-						<div class="back-to-module">
+						<div class$="{{_getDefaultClass('back-to-module')}}">
```
ChrisKraljevicD2L, line 148: nit: does this have to be a two-way binding? i.e. could `{{...}}` be switched out for `[[...]]`?

d2l-sequence-viewer/d2l-sequence-viewer-header.js
```diff
@@ -139,16 +145,16 @@ PolymerElement) {
...
 							<slot name="d2l-back-to-module"></slot>
 						</div>
 						<template is="dom-if" if="[[!isSingleTopicView]]">
-							<div class="flyout-menu">
-								<d2l-icon class="flyout-divider" icon="d2l-tier2:divider-big"></d2l-icon>
-							</div>
-							<div class="d2l-flyout-menu">
-								<slot name="d2l-flyout-menu" d2l-flyout-menu=""></slot>
-							</div>
+								<div class$="{{_getDefaultClass('flyout-menu')}}">
```
ChrisKraljevicD2L, line 152: same about the two-way binding here

d2l-sequence-viewer/d2l-sequence-viewer-header.js
coffeexcode: I think because it's a function call that it is required. I wasn't able to get it working with `[[...]]` as is, but maybe I can see if making it a property works for 1-way binding

ChrisKraljevicD2L: Yeah computed properties can be used in one-way bindings. Functions get kinda funky. But I understand not wanting to put more properties in when this is already being computed with an existing public property.

I'm trying to find where I read this, but I recall two-way bindings being deprecated. The favoured practice is one-way bindings putting data in the DOM and passing it into child elements, and events passing info back up to the host.

idk maybe I just like square brackets more :P

ChrisKraljevicD2L: either if this is working, please don't let this hold up the PR from getting merged before branching. I'll slap a ✔️ on this

ChrisKraljevicD2L (approved):

brissons (approved):

PR #343, opened 2021-11-23
DE46051: LX > Expand/collapse button in wrong position after expanding

prajapac:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F614823232359

## Description
- The flyout menu buttons were not transitioning upon expanding/collapsing of the menu after changes made during this PR: https://github.com/BrightspaceHypermediaComponents/sequences/pull/340/files

## Goal/Solution
- The changes made here fix the transitioning defect by adding an non-kbd accessible invisible div that resizes appropriately to allow for the buttons to be in the correct spot as well as work with transitioning.
- Further, (after checking with designer), the CSS was also modified to allow for uniformity regardless of what label text says, i.e. `Back` vs. `Back to Content`

## Video/Screenshots

Before:
1) Back
back_before

2) Back to Content
back_to_content_before

After:
1) Back
back_after

2) Back to Content
back_to_content_after


coffeexcode (approved):
LGTM

d2l-sequence-viewer/d2l-sequence-viewer-header.js
```diff
@@ -79,29 +79,35 @@ PolymerElement) {
...
 				padding-left: initial;;
 				padding-right: var(--d2l-sequence-viewer-button-padding);
 				margin-left: initial;
 				margin-right: 0;
 			}
+			:host([is-sidebar-closed]) .flyout-menu {
+				display: flex;
+				flex-direction: row;
+				align-items: center;
+				padding-left: 18px;
```
brissons, line 103: do we need to add padding to the rtl version for `flyout-menu` as well? Probably something like:

```
padding-right: 18px;
padding-left: initial;
```

d2l-sequence-viewer/d2l-sequence-viewer-header.js
```diff
@@ -112,6 +118,16 @@ PolymerElement) {
 			:host(:dir(rtl)) .d2l-flyout-menu {
 				padding: 0 var(--d2l-flyout-menu-left-padding) 0 var(--d2l-flyout-menu-right-padding);
 			}
+			.d2l-header-left-resize {
```
brissons, line 121: nit: I would maybe rename the class to not be left specific due to rtl

brissons (commented):
Will do a quick dev test, but code lgtm with a couple comments:

ChrisKraljevicD2L (approved):

ChrisKraljevicD2L (changes requested):
Actually, Scott's review comments are important to address. We don't want to break RTL.

brissons (commented):
Dev. test looks good!

Small nit: I noticed that the slide out transition starts further to the right when the text is set to 'Back' (might be hard to notice):

back-example
It looks like the transition starts at the end of `header-left-inner`, which is set to `max-width: 260px;`. Changing this value changes where the transition starts.

ChrisKraljevicD2L (approved):

d2l-sequence-viewer/d2l-sequence-viewer-header.js
prajapac: Addressed in most recent commit!

d2l-sequence-viewer/d2l-sequence-viewer-header.js
prajapac: Addressed in most recent commit!

prajapac:
> Dev. test looks good!
> 
> Small nit: I noticed that the slide out transition starts further to the right when the text is set to 'Back' (might be hard to notice):
> 
> back-example It looks like the transition starts at the end of `header-left-inner`, which is set to `max-width: 260px;`. Changing this value changes where the transition starts.

I'll log a defect for this one, since it seems to be a different/pre-existent issue.

brissons (approved):

prajapac:
> > Dev. test looks good!
> > Small nit: I noticed that the slide out transition starts further to the right when the text is set to 'Back' (might be hard to notice):
> > back-example It looks like the transition starts at the end of `header-left-inner`, which is set to `max-width: 260px;`. Changing this value changes where the transition starts.
> 
> I'll log a defect for this one, since it seems to be a different/pre-existent issue.

https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F617141761093

PR #344, opened 2021-11-23
DE45494 Prevent offscreen interaction with side nav

ChrisKraljevicD2L:
## Relevant Rally Links

- [DE45494](https://rally1.rallydev.com/#/289692574792/dashboard?detail=%2Fdefect%2F610814189091)

## Description

Screenreaders and keyboard navigation would interact with the side nav even collapsed and off-screen.

## Goal/Solution

Applying the CSS rule `visibility: hidden`  both prevents screenreader output as well as tabbing to an element or any of its children.

## Video/Screenshots

before

you can see there's a large time gap between the "next" button highlighting" and the main content panel highlighting. I was still pressing tab with the same cadence as before between those two. This shows that keyboard navigation is indeed going through the side nav.

after

That time gap is now gone, no more tabbing through stuff off-screen.

## Related PRs

- N/A

## Remaining Work


d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -756,6 +760,9 @@ class D2LSequenceViewer extends mixinBehaviors([
 
 		if (!this.isSidebarClosed) {
 			sidebarContainer.classList.remove('hide-overflow');
+			
+		} else {
+			sidebarContainer.classList.add('prevent-offscreen-interaction');
```
ChrisKraljevicD2L, line 765: There is a reason why the addition of the `prevent-offscreen-interaction` class is in this transitionend listener function and now in `_onSidebarClose`.

`visibility: hidden` causes an element's width to drop to 0 immediately, something we cannot transition on. To maintain the nice open/close animation, I only add this class after the animation has completed.

ianlaidlaw (approved):

kristian-torres (approved):

PR #345, opened 2021-11-24
Hide completion counts when they would show 0/0

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F613009287361&fdp=true

## Description

If the completion count should ever show `0/0`, we want to hide that from the user.

## Video/Screenshots

# Before:

## Launcher:
fulllauncher

## Sidebar - new progress bar
fullsidebar1

## Sidebar - old progress bar
fullsidebar2

## End of Unit Page
fulleou

# After:

## Launcher: 
emptylauncher

## Sidebar
emptysidebar

## End of Unit Page
emptyeou

PR #349, opened 2021-11-25
Hide completion counts when they would show 0/0 [Followup]

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F613009287361&fdp=true

## Description

This is a follow up to [my previous pr](https://github.com/BrightspaceHypermediaComponents/sequences/pull/345) which missed some spec. It fixes the following things:
1. Unrelated, but Lesson completion counts were not showing due to a previous defect. This fixes that and then hides them if the total is `0/0`
2. Hides `0/0` completion counts for folders

I also went through with a screen reader and ensured that `0/0` completion counts aren't being unexpectedly read out when they would be visually hidden.

### BEFORE:

image

### AFTER:

image


d2l-sequence-navigator/d2l-outer-module.js
```diff
@@ -601,7 +604,7 @@ class D2LOuterModule extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 	}
 
 	_hasChildren(entity) {
-		return entity && entity.getSubEntities().length !== 0;
+		return HypermediaHelper.getTopicAndModuleSubEntities(entity).length !== 0;
```
Kiefer-Co-D2L, line 607: Question: What's the purpose of this change?

ianlaidlaw: This was the "unrelated" bug I talked about in the description. Basically at some point in the past, another team changed how `EntityBehaviour` worked. If you passed in nothing to `getSubEntities()` it used to mean "give me ALL subentities, regardless of the rel." but they changed it to require a rel or a regexp. When we fixed this bug, I made this helper function that does exactly that. This is just a missed spot that was never updated. Thankfully it was only visual.

ianlaidlaw: https://github.com/BrightspaceHypermediaComponents/sequences/pull/317

Kiefer-Co-D2L (approved):

PR #350, opened 2021-11-25
Hide completion checkmark for Optional and Exempt activities

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F613009291913&fdp=true

## Description

Previously, exempt and optional topics would show checkmarks next to them. We want to hide this, even if the topic is completed.

## Video/Screenshots

# Before:

image
image

# After:

image
image


d2l-sequence-navigator/d2l-activity-link.js
```diff
@@ -266,10 +266,6 @@ class D2LActivityLink extends PolymerASVLaunchMixin(CompletionStatusMixin()) {
 				type: Boolean,
 				computed: '_hasIcon(entity)'
 			},
-			completionStatus: {
```
ianlaidlaw, line 269: as far as i could possibly find, this was not used anywhere. seemed safe to delete

d2l-sequence-navigator/d2l-completion-status.js
```diff
@@ -29,32 +29,17 @@ class D2LCompletionStatus extends CompletionStatusMixin() {
 			completionStatus: {
 				type: String,
 				computed: '_getCompletionStatus(entity)',
-				observer: '_showCompletionType'
+				observer: '_getShowCompletionIcon'
 			},
-			isOptional: {
```
ianlaidlaw, line 34: was not used at all. seemed to be copy+paste from `d2l-completion-requirement.js` which is similar but slightly different

mixins/completion-status-mixin.js
```diff
@@ -92,6 +92,12 @@ export function CompletionStatusMixin() {
 				return;
 			}
 
+			if (subEntity
```
ianlaidlaw, line 95: this was safe to add as this function was only used in the one spot i was expecting it to be used in. nothing else depends on this

d2l-sequence-navigator/d2l-completion-status.js
```diff
@@ -15,7 +15,7 @@ class D2LCompletionStatus extends CompletionStatusMixin() {
 				color: inherit;
 			}
 		</style>
-		<template is="dom-if" if="[[isComplete]]">
+		<template is="dom-if" if="[[_showCompletionIcon]]">
```
ianlaidlaw, line 18: i renamed this because `isComplete` has pretty strong meanings, but it's moreso just a visual thing

coffeexcode (approved):
LGTM. Nice on the cleanup while you we're making the change 👍

Kiefer-Co-D2L (approved):
Looks good!

PR #356, opened 2021-12-03
DE46171 > Menu expand transitions starts too far to the right

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F617141761093

## Description

> As part of work done on this defect: https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F614823232359, it seems that there is a max-width CSS value for the 'header-left-inner':
>
> image (7)
>
> This value determines where the buttons start transitioning from, and seems to be cherry-picked for when the label text says "Back to Content". Since that text now sometimes says "Back" and can also be overridden due to OSLO, the value should be changed to be calculated dynamically so the transition is smooth regardless of what the label text says.
>
> See below for visual:
>
> 1. Back to Content (Slowed):
>
> back_to_content (1)
>
> 2. Back (Slowed):
>
>back (2)

## Goal/Solution

The sliding sidebar transition is based off of the `#header-left-inner`'s *max-width*, which is a hardcoded value that depends on whether the sidebar is open or closed.  The closed value is set such that the animation looks correct only when the text says "Back to Content".

The solution is to make *max-width* when closed a function of `.d2l-navigation-link-back-text`'s size.

Note that there is responsive behaviour which is overridden, but which correctly looks smooth (when screen width is very small, the back text is absent, and the new behaviour will start sliding at a lower value as expected):

```css
			@media(max-width: 435px) {
				:host([is-sidebar-closed]) #header-left-inner {
					max-width: 130px;
				}
			}
```

## Video/Screenshots

Here's the fixed behaviour seen with various screen widths:

Smooth Sidebar Transition

For reference, here is another way to replicate the problem in real-time.  It's very brief, but the hamburger menu/close button skips to where "Back to Content" would end even when the text is different:

Not Smooth Sidebar Transition 2

## Remaining Work

- [X] Handle W.R.T. screen width


d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -689,13 +692,26 @@ class D2LSequenceViewer extends mixinBehaviors([
 		this._moduleProperties = properties;
 	}
 
+	_sideBarSetInnerHeaderWidth(width = null) {
+		const headerLeftInner = document.getElementsByTagName('d2l-sequence-viewer')[0].shadowRoot.querySelector('d2l-sequence-viewer-header').shadowRoot.querySelector('#header-left-inner');
+
+		if (width) {
+			headerLeftInner.style.maxWidth = width + 'px';
+		} else {
+			const transitionWidth = 152;
```
Kiefer-Co-D2L, line 701: Is it okay to have magic numbers like this or should they be made class-level?  I see a similar usage on line 755.

ianlaidlaw: not ideal, but sometimes there isn't a great way around it easily. I think at the very least (and the same should have been true for 755) is to at least name it and put it in a variable

Kiefer-Co-D2L: I think I know a good name for the other one.  Will at least be a bit more explicit

ianlaidlaw:
i'll give this one a dev test

ianlaidlaw (approved):
dev tested, LGTM. fixes the issue where the sidenav opened on changing resolutions, awesome

d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -689,13 +693,27 @@ class D2LSequenceViewer extends mixinBehaviors([
...
+			const navigationLinkBackText = document.getElementsByTagName('d2l-sequence-viewer')[0].shadowRoot.querySelector('d2l-navigation-link-back').shadowRoot.querySelector('.d2l-navigation-link-back-text');
+			headerLeftInner.style.maxWidth = (navigationLinkBackText.offsetWidth + transitionWidth) + 'px';
+		}
+	}
+
 	_sideBarOpen() {
 		const sidebarContainer = this.shadowRoot.getElementById('sidebar-container');
 		const viewframeFogOfWar = this.shadowRoot.getElementById('viewframe-fog-of-war');
+		const sidebarOpenWidth = 570;
```
prajapac, line 711: How did you obtain the values 152 and 570?

Kiefer-Co-D2L: They're from the styles in this file:
`sequences\checkout\d2l-sequence-viewer\d2l-sequence-viewer-header.js`

From my understanding, the transition is meant to start right next to the back text and grow to 570px.
Initially, it was hardcoded as 260px.

I got 152 by subtracting the offset width of `.d2l-navigation-link-back-text` from 260.
When you add 152 to the back text element's width, you should get a transition that looks smooth.
E,g, in the gif, closed width is now 188px when the text says "Back", and still 260px when the text says "Back to Content".  The open width will always be 570px.

The only issue I foresee is if the text is long enough to exceed 570px, but that was a problem with the original code too.  I'm not sure if any OSLO translations could cause an issue, but the translation would have to be >4 times as long as "Back to Content" in English.

prajapac (approved):

PR #362, opened 2021-12-08
Fix progress wheel on smaller resolutions

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F619090941737&fdp=true

## Description

The progress wheel looks comically small on lower resolution screens.

## Goal/Solution

Give it a min width so that it doesn't get too small, but still respects the same width on higher resolution screens. I chose the width kind of arbitrarily, but I wanted to make sure at the very least we could fit a 2-digit numerator and a 2-digit denominator comfortably inside the circle at the smallest size and still be able to read it.

## Video/Screenshots

Ignore the data in these screen shots, I just fudged some data to get this view to show up.

# BEFORE:

image

# AFTER:

image


components/d2l-sequences-end-of-lesson-progress.js
```diff
@@ -12,6 +12,7 @@ class D2LSequencesEndOfLessonsProgress extends ASVFocusWithinMixin(CompletionSta
 			}
 			d2l-meter-circle {
 				width: 12%;
+				min-width: 80px;
```
ianlaidlaw, line 15: I mentioned this in the pr description, it's kind of an arbitrary number but a 2-digit numerator and a 2 digit denominator fit comfortably inside this width.

Kiefer-Co-D2L: Would it be worth the effort to have it be larger for bigger counts?  Otherwise LGTM!

ianlaidlaw: I mistakenly thought the text didn't scale when I made this comment at first, so no matter the size a bigger count will look weird. That may be out of the scope of this defect, but I could look into making that look better. give the text a max-size or something. because right now 3digits out of 3digits will always overflow

ianlaidlaw: image

this is master

brissons (approved):

components/d2l-sequences-end-of-lesson-progress.js
ianlaidlaw: i'm gonna try and fix the text also as part of this pr

ianlaidlaw: nevermind, it's a core component. if i fix it, it would be in a future pr

PR #366, opened 2021-12-15
Add missing padding below certain elements

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F617349674219

## Description

Some whitespace looks off, especially when units/lessons/folders are empty.

## Goal/Solution

Whitespace was added after certain elements such that they appear to have the same amount of space above and below.
See screenshots for more detail!

## Video/Screenshots

Skeletons and hover effect have been turned on and darkened for visibility.

| Before | After | Change |
| - | - | - |
| Empty Unit Old | Empty Unit New | When a unit is empty, the launch unit button was touching the bottom of the sequence launcher accordion.  For non-empty units, this is not the case as the module list has padding and margin above it.  To fix this, an empty div appears only if the unit is empty.  This adds 15px of space, matching what's above the button container.  |
| Empty Lesson Old | Empty Lesson New | When a lesson is empty, the button or skeleton appears to have less whitespace below it than above it.  To fix this, an empty div appears only if the outer module (lesson) has no children.  This adds 6px of space, which matches the padding normally present above module contents.  |
| Empty Folder Old | Empty Folder New | This is something I noticed but isn't in the ticket.  There's no whitespace after the last item in a lesson.  This affects both empty and non-empty folders and is fixed by adding padding only after the last list element. |

For sanity, here is a non empty unit, lesson, and folder to show no unnecessary whitespace:

<p align="center">
Nothing is broken!
</p>

## Related PRs

- Sequence Launcher (sequences must be merged first): https://github.com/Brightspace/sequence-launcher-fra/pull/225

## Remaining Work


ianlaidlaw (approved):

ChrisKraljevicD2L (approved):

ChrisKraljevicD2L:
I'll give this a quick test

ChrisKraljevicD2L (approved):
Dev tested :+1:

PR #367, opened 2021-12-16
Fix missing/incorrect imports

devpow112:
## Relevant Rally Links

None.

## Description

Fix missing/incorrect imports. Both `d2l-colors` and `d2l-offscreen` were used but not actually referenced as dependencies. Switch to `@brightspace-ui/core` import paths (since they are drop in replacements) and fix missing deprecated color. Also add explicit use of `d2l-link` since this can't be switched to `core` right now.  These worked since same upstream dependency is pulling them in. If that changed then these would be missing.

## Goal/Solution

Fix incorrect imports by mapping them to `@brightspace-ui/core` equivilents.

## Video/Screenshots

None.

## Related PRs

None.

## Remaining Work

None.


d2l-sequence-viewer/d2l-sequence-viewer-new-content-alert.js
```diff
@@ -42,7 +42,7 @@ class D2LSequenceViewerNewContentAlert extends mixinBehaviors([
 					#new-content-alert {
 						--d2l-alert-icon-dimensions: 18px;
 						background: white;
-						border: 1px solid var(--d2l-color-titanius);
+						border: 1px solid var(--d2l-color-mica);
```
devpow112, line 45: the `--d2l-color-titanius` is deprecated and just maps to `--d2l-color-mica`.

dlockhart (approved):

PR #371, opened 2022-01-05
Update html-block API usage

grant-cleary:
See this PR for context: https://github.com/BrightspaceUI/core/pull/1960

Slight change to the html-block API, in that it now takes its contents directly in a liteDOM slot vs. within a template. Otherwise behaviour should remain the same.


brissons (approved):
thanks Grant! ran a quick smoke test and lgtm.

PR #378, opened 2022-01-18
Reorganize file structure and remove outdated demos

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F622544966929&fdp=true

## Description

This repo previously used demos for testing. These became very out of date, sometimes broken, and more or less unneeded as the ability to run BSI locally was introduced, which made testing overall much easier. This also reorganizes some files instead of just having a bunch of components sitting in a top level directory. There should be no functional change in sequences after these changes.

### Before:

image

### After:

image


ianlaidlaw:
I will be incrementing the major version as part of this pr

ianlaidlaw:
The easiest way to review this pr is probably to check it out, make sure it works, and look at the file structure in an IDE

components/end-of-lesson/d2l-sequences-end-of-lesson-progress.js
```diff
@@ -1,5 +1,5 @@
-import { CompletionStatusMixin } from '../mixins/completion-status-mixin.js';
```
brissons, line 1: up to you, but I noticed the naming of the component doesn't follow the convention of the other end-of-lesson components. should it be something like `d2l-sequences-content-eol-progress-meter`?

brissons (commented):
lgtm! definitely a lot cleaner. I played around with it locally and it looks like its working as expected. Just one small comment:

brissons (approved):

components/end-of-lesson/d2l-sequences-end-of-lesson-progress.js
ianlaidlaw: i noticed that too! i think i meant to rename it early on but i forgot about it. i will rename it to make it consistent, thanks!

PR #353, opened 2021-12-01
Create Total Activities Component for End of Unit pages

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/dashboard?detail=%2Fuserstory%2F617359155011&fdp=true

## Description

This pr creates a new component that will display how many of required or optional topics are in a unit. This component is not in use yet, but will be as part of the end of unit redesigns. There exists a subEntity on the "end of sequence" entity that contains these numbers and will spit them out.

This component will only render if any optional topics exist, per the design.

## Video/Screenshots

image

image

image

## Related PRs

- lms:
- sequences:

PR #382, opened 2022-01-27
Convert a component to Lit Html

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fuserstory%2F622544965323&fdp=true

## Description

The goal of this pr is to convert a single component from polymer 3 to lit, and introduce us the ability to slowly replace all old components. There are 2 main differences with these new components that had cascading changes to the entire project:

1. Lit components will use the siren-sdk entity stuff, much like we already do in sequences. This means defining an Entity type (sequences already has some stuff) and then accessing things via that representation, rather than doing things directly.
2. Lit components use the LocalizeDynamicMixin from core. This had a cascading effect where I converted all lang files to `.js` files and deleted a ton of useless code that was simply turning our langs from individual json files -> locales.json -> localize-behavior.js using a handlebars template. All of that is gone now. This means that any changes should just reflect automatically, no generation required, even for old Polymer components.

The component I converted was [the end-of-sequence-summary componen](https://github.com/BrightspaceHypermediaComponents/sequences/pull/353)t. It seemed like a good candidate as it was recent, simple, has locales, and conditional rendering.

I tried to post lost of comments, some important files are collapsed due to large changes. You can ignore all the changes to lang files, as those are simply converting them from .json to .js files.

The most important files to be sure to read are:
[d2l-sequences-content-summary-lit](https://github.com/BrightspaceHypermediaComponents/sequences/pull/382/files#diff-502ff9c5d1c9661d81818e10b5963c983b338d3ff5f882299d735485584193f8R1)
[localize-behavior](https://github.com/BrightspaceHypermediaComponents/sequences/pull/382/files#diff-ce8ab7c01d50948319beb9bf5fcbadf10a4f0bbc9c87c073bf222ddff40a4852L1)

## Video/Screenshots

image

image

## Related PRs

- siren-sdk: https://github.com/BrightspaceHypermediaComponents/siren-sdk/pull/449

## Stuff left to test more thoroughly
- [ ] how this interacts with sequence launcher FRA
- [x] investigate "harmless" 404 on page load for lang file?


.eslintrc.json
```diff
@@ -18,5 +18,14 @@
 		"D2LSequencesContentEoLMain": false,
 		"D2LSequencesContentModule": false,
 		"D2LSequencesContentError": false
+	},
+	"rules": {
```
ianlaidlaw, line 22: updating eslint and everything to work with Lit was producing a lot of errors. instead of making this pr even messier, im just going to supress them for now and we can clean them up in a followup pr

components/end-of-lesson/d2l-sequences-content-eol-in-progress.js
```diff
@@ -82,10 +82,10 @@ export class D2LSequencesContentEoLInProgress extends D2L.Polymer.Mixins.Sequenc
 							token="[[token]]">
 						</d2l-missed-activity>
 
-						<d2l-sequences-content-eol-summary
-							href="{{href}}"
+						<d2l-sequences-content-eol-summary-lit
+							href="[[href]]"
```
ianlaidlaw, line 86: no more 2 way data binding

components/end-of-lesson/d2l-sequences-content-eol-in-progress.js
```diff
@@ -5,7 +5,7 @@ import './d2l-sequences-content-eol-progress.js';
 import './d2l-sequences-content-eol-missed.js';
 import '@polymer/polymer/polymer-legacy.js';
 import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
-import './d2l-sequences-content-eol-summary.js';
+import './d2l-sequences-content-eol-summary-lit.js';
```
ianlaidlaw, line 8: i kept the `-lit` just for more destinction, but im down to remove it if people want

localize-behavior.js
```diff
@@ -1,58 +1,128 @@
-import '@polymer/polymer/polymer-legacy.js';
```
ianlaidlaw: github is auto collapsing this file, but please read it as it's very important

localize-behavior.js
```diff
@@ -1,58 +1,128 @@
...
+	}
+};
+
+/** @polymerBehavior D2L.PolymerBehaviors.Sequences.LocalizeBehavior */
+D2L.PolymerBehaviors.Sequences.LocalizeBehavior = [
+	D2L.PolymerBehaviors.LocalizeBehavior,
+	D2L.PolymerBehaviors.Sequences.LocalizeBehaviorImpl
+];
+
+export const SequencesLocalizeMixin = superclass => class extends LocalizeDynamicMixin(superclass) {
```
ianlaidlaw: this mixin does all the localization/oslo for Lit components

localize-behavior.js
```diff
@@ -1,58 +1,128 @@
...
+ */
+D2L.PolymerBehaviors.Sequences.LocalizeBehaviorImpl = {
+	properties: {
+		/**
+		 * Localization resources.
+		 */
+		resources: {
+			value: function() {
+				return {
+					'ar': ar,
```
ianlaidlaw: these are just brought in direction, for some reason this had like 3 layers of middleware to get the terms from the json files to here. in the future, when we get new terms they should just automatically work

plop/localize-behavior.hbs
```diff
@@ -1,58 +0,0 @@
-import '@polymer/polymer/polymer-legacy.js';
```
ianlaidlaw, line 1: file no longer needed... pointless middleware

test/d2l-sequence-navigator/d2l-sequence-navigator-activity.html
```diff
@@ -27,7 +27,6 @@
 import '../../d2l-sequence-navigator/d2l-sequence-navigator.js';
 import SirenFixture from 'polymer-siren-test-helpers';
 
-/* global SirenFixture */
```
ianlaidlaw, line 30: eslint error, supressing/rules didnt work

test/data/d2l-nav.js
```diff
@@ -1,102 +1,102 @@
-/*
```
ianlaidlaw, line 1: this file is not new. again, eslint rules/supression wasnt working here so i just fixed it (line endings)

test/data/multipage1.html
```diff
@@ -1,14 +1,15 @@
-<html id="page1">
```
ianlaidlaw, line 1: same thing, eslint

components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js
```diff
@@ -0,0 +1,65 @@
...
+
+class D2lSequencesContentEoLSummaryLit extends SequencesLocalizeMixin(EntityMixinLit(LitElement)) {
+	static get properties() {
+		return {
+			href: {
+				type: String
+			},
+			token: {
+				type: String
+			}
```
ChrisKraljevicD2L: Do these need to be declared here? I think they get brought in with `EntityMixinLit`

package.json
```diff
@@ -25,22 +22,24 @@
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@babel/core": "^7",
```
ianlaidlaw, line 25: not sure if i actually need this or not, will confirm

localize-behavior.js
```diff
@@ -1,58 +1,128 @@
...
+/** @polymerBehavior D2L.PolymerBehaviors.Sequences.LocalizeBehavior */
+D2L.PolymerBehaviors.Sequences.LocalizeBehavior = [
+	D2L.PolymerBehaviors.LocalizeBehavior,
+	D2L.PolymerBehaviors.Sequences.LocalizeBehaviorImpl
+];
+
+export const SequencesLocalizeMixin = superclass => class extends LocalizeDynamicMixin(superclass) {
+	static get localizeConfig() {
+		return {
+			importFunc: async lang => (await import(`./lang/${lang}.js`)).default,
```
ianlaidlaw: on first load, this import is giving a 404, but then it immediately seems to resolves itself. i am investigating why that is

ChrisKraljevicD2L (approved):
LGTM :+1:

components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js
ianlaidlaw: nah u right, no real need to declare them

components/end-of-lesson/d2l-sequences-content-eol-in-progress.js
brissons: I think it makes sense for now! We will probably want to change it back when we start migrating more components (so not everything has `-lit`)?

localize-behavior.js
brissons: soooo nice!

brissons (approved):
lgtm!

localize-behavior.js
ianlaidlaw: so it's looking for en-us which doesn't exist, but then seems to fall back to en which is making this work.. not too sure what exactly is going on here. other repos dont have an en-us file either

ianlaidlaw: it seems harmless so i'm going to leave it for now

brissons: It looks like the mixin fallbacks to English if the culture/lang code can't be found:
https://github.com/BrightspaceUI/core/blob/72d2b7605180505f1405c7ecf64cc4da66220ebd/mixins/localize-dynamic-mixin.js#L4

components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js
```diff
@@ -0,0 +1,61 @@
...
+			href: { type: String },
+			token: { type: String }
+		};
+	}
+
+	static get styles() {
+		return [
+			css`
+				:host {
+					@apply --d2l-body-standard-text;
```
svanherk, line 11: Is this working?  I wonder if somewhere else is importing the sass stuff. For Lit components, I think we want to set things up like this:
```
import { bodyStandardStyles } from '@brightspace-ui/core/components/typography/styles.js';
...
	static get styles() {
		return [ bodyStandardStyles ,
			css`
...
		return html`
			<div class="d2l-body-standard">
...

File: localize-behavior.js (hunk 13 above)
Comment (svanherk) [reply localize-behavior.js:124]: So this is a known issue for dev only: https://github.com/BrightspaceUI/core/issues/1544
Hopefully fixed soon!  I don't think you need to worry about it

Comment (ianlaidlaw) [reply localize-behavior.js:124]: right, so that explains why this works but i wonder if it's by design that it should fail for en-us by default first, the other repos i've been looking at also don't have those lang files, just en.js from what i've seen. i'll do some testing with those actual projects and see if those are also 404'ing

File: components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js (hunk 14 above)
Comment (ianlaidlaw) [reply components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js:11]: that's a good question. when i recreated this component i more or less copy and pasted all the css from the old one and compared them to make sure they looked the exact same - which they did, and then didn't think too much of it after that. but it's entirely possible that this line isn't what is making that be the case. i'll investigate and see if this is needed or not

File: localize-behavior.js (hunk 13 above)
Comment (svanherk) [reply localize-behavior.js:124]: IIRC, the fallback order is correct, and the prod builds will know not to check for `en-us` because it doesn't exist and doesn't get added to the switch statement.  But for our dev builds that aren't built with rollup, it tries the whole path.  We have https://github.com/BrightspaceUI/core/pull/1576 up to address, we just need to get back to it

File: components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js (hunk 14 above)
Comment (ianlaidlaw) [reply components/end-of-lesson/d2l-sequences-content-eol-summary-lit.js:11]: you're right, this isn't doing anything but it still looks the same without it, so something must be happening behind the scenes. i'm going to remove it from this component for this pr, and make a note to investigate the other components in this repo and see what is actually happening.

File: d2l-sequences.serge.json [hunk 15]
```diff
@@ -1,7 +1,7 @@
 {
     "name": "d2l-sequences",
     "source_dir": "lang",
-    "output_file_path": "lang/%LANG%.json",
```
svanherk, line 4: So I'm not sure because I don't know a ton about serge, but for other repos that usually use the `js` files for serge they also have this in the config:
```
  "parser_plugin": {
    "plugin": "parse_js"
  },
  "source_match": "en\\.js$",
```
Have you seen somewhere that works without these?  I wonder if they're the default?

localize-behavior.js
```diff
@@ -1,58 +1,128 @@
...
-D2L.PolymerBehaviors.Sequences.LocalizeBehavior = [
-	D2L.PolymerBehaviors.LocalizeBehavior,
-	D2L.PolymerBehaviors.Sequences.LocalizeBehaviorImpl
-];
+import '@polymer/polymer/polymer-legacy.js';
+import 'd2l-localize-behavior/d2l-localize-behavior.js';
+import { getLocalizeOverrideResources } from '@brightspace-ui/core/helpers/getLocalizeResources.js';
+import { LocalizeDynamicMixin } from '@brightspace-ui/core/mixins/localize-dynamic-mixin.js';
+
+import ar from './lang/ar.js';
```
svanherk: Probably easier to deal with in another PR, but I don't think all of these lang files are needed anymore.  I believe serge only provides updates for these languages: https://github.com/Brightspace/serge-localize/blob/master/static/fra.common.serge.inc#L7
And rewrites them to what's in your serge `output_lang_rewrite` (or the default https://github.com/Brightspace/serge-localize/blob/master/static/fra.common.serge.inc#L20 if you have no `output_lang_rewrite` entry).

So I think a lot of these files will always be exactly the same, and our localization libraries should handle falling back appropriately.  If there's no differences between the specific root language and the region one, I think all you need is:
`ar.js`, `cy.js`, `da.js`, `de.js`, `en.js`, `es.js`, `es-es.js`, `fr.js`, `fr-fr.js`, `fr-on.js`, `ja.js`, `ko.js`, `nl.js`, `pt.js`, `sv.js`, `tr.js`, `zh-cn.js`, `zh-tw.js`

There's a bit more info [here](https://d2l.slack.com/archives/C0PHG3QB0/p1638464485221600), but I can help if you want to get these cleaned up!

svanherk (approved):
Looks good!  I'd just look into that serge file, but it's a bit hard to test without waiting for serge to run again in a month.

localize-behavior.js
ianlaidlaw: thats great, i'll definitely look into cleaning this up in a future pr. i will absolutely reach out for help if i need it.

d2l-sequences.serge.json
ianlaidlaw: i think you're right. yep i wasnt sure the best way to verify this yet until the next time the serge integration happened.. but i haven't seen any repos that don't have these, so i'm going to add them

PR #406, opened 2022-03-03
DE42056 Google Slides Issues

Kiefer-Co-D2L:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F483764895804

## Description

Issue 1:

> We have had trouble with google slides all year. Any time I create a link to a slideshow, it ignores the "open in new window" function.

Issue 2:

> The embedded slideshow then breaks when students go to scroll through the slides. The screen will duplicate and freeze. When I try to impersonate a student to view the slideshow, I get a message that BrightSpace cannot connect with google at that particular moment. My students have had to quickly make a copy of the slideshow before the screen freezes.

> Also noticing that Google slides don't render correctly at all, the toolbar is in the middle of the one slide, see attachment frozen slides screenshot.png

### Open in New Window Being Ignored

~This issue does not appear to replicate on quad or sitelord.  On my local instance, Google Slides topics cannot be created.~

~This PR adds some logging that might shed light on the issue if it reappears: https://github.com/Brightspace/lms/pull/20030~

~A possible reason for the opening in new tab setting not appearing is the "open-in-new-tab" class being absent, which might occur if `targetFrame` is incorrect on the `topicInfo`.~

This issue appears to be solved by a previous PR: https://github.com/BrightspaceHypermediaComponents/sequences/pull/391

### Google Slides Rendering Incorrectly

This appears to be due to multiple iframes rendering as children of the [`polymer-google-drive-viewer`](https://github.com/Brightspace/polymer-google-drive-viewer/blob/master/google-drive-viewer.js).

When multiple Google Slides iframes are rendered at once, two iframes may be placed one above each other such that the toolbar of the second iframe appears in the middle of the screen.  This behaviour is pictured below.

In correct operation, new iframes should be periodically created and the old ones closed.  This is not the case, as instead all iframes are immediately "closed" by the viewer (but the elements are not removed).  To remedy this, all iframes except for the latest are closed whenever a new iframe appears.  As the last child of the shadowroot within `google-drive-viewer` is latest created, this mimics correct behaviour.

## Video/Screenshots

Rendering issue:

Toolbar in Middle

The last child iframe is the youngest by timestamp, seen here as id, and therefore the only child to be spared:

image

The "open-in-new-tab" class appears to be being added correctly.  `_blank` is one of the possible values of `targetFrame` that trigger the class being added:

image

## Related PRs

- ~https://github.com/Brightspace/lms/pull/20030~
- https://github.com/BrightspaceHypermediaComponents/sequences/pull/391

## Remaining Work


components/content-viewers/d2l-sequences-content-link-googledrive.js
```diff
@@ -44,5 +50,33 @@ export class D2LSequencesContentLinkGoogledrive extends D2L.Polymer.Mixins.Seque
 			return '';
 		}
 	}
+
+	_initIFramesObserver() {
+		const self = this;
+		const viewer = this.shadowRoot &&
+		this.shadowRoot.querySelector('google-drive-viewer') &&
+		this.shadowRoot.querySelector('google-drive-viewer').shadowRoot;
```
kristian-torres, line 58: if, for some reason, there is no google-drive-viewer shadowRoot, your observer will fail..

components/content-viewers/d2l-sequences-content-link-googledrive.js
```diff
@@ -44,5 +50,33 @@ export class D2LSequencesContentLinkGoogledrive extends D2L.Polymer.Mixins.Seque
...
+					self._clearIFrames(viewer);
+				}
+			}
+		};
+
+		const observer = new MutationObserver(callback);
+		observer.observe(viewer, {childList: true});
+	}
+
+	_clearIFrames(viewer) {
```
kristian-torres, line 72: Is there a way to leave some sort of identifier on the specific iframe you are searching for? This way we can constantly update the google-drive-viewer's shadow dom by removing the latest iframe and add the new one instead of removing all iframes except for the last one?.

If that's not possible how about we take the base case of the first iframe that's loaded and constantly update it's parent to remove the iframe in question and replace it when necessary.

Does that sound like a reasonable option?

components/content-viewers/d2l-sequences-content-link-googledrive.js
ChrisKraljevicD2L: I'm trying to find some kind of lifecycle hook that would allow us to re-check whether or not `<google-drive-viewer>`'s `shadowRoot` exists. There isn't a nice `render` function in polymer like we have in Lit. And putting this logic in the `template` get function is a baaaaaaaaad idea.

I think this is ok for now, since the `ready` lifecycle callback is pretty late in polymer's setup. google-drive-viewer should have its shadow dom initialized at roughly the same time. Not the best but I can't think of anything better atm.

components/content-viewers/d2l-sequences-content-link-googledrive.js
Kiefer-Co-D2L: It would be possible to add an id to the iframes, but it would have to be in [this repo](https://github.com/Brightspace/polymer-google-drive-viewer/blob/master/google-drive-viewer.js).

I might be misunderstanding but what's the benefit of re-adding the iframe?  Adding and removing them looks like it should be handled in `google-drive-viewer`, but it isn't behaving as expected so this is a workaround in sequences.

components/content-viewers/d2l-sequences-content-link-googledrive.js
Kiefer-Co-D2L: Thanks, I'll move getting this viewer to the `_clearIFrames` function itself!

kristian-torres: at the very least, we'd need to run an initialized check on the viewer but as Chris said, there's no nice out of the box functionality (dammit polymer) to check the lifecycle of the shadowRoot. :(

PR #409, opened 2022-03-08
import hindi

GurpinderSandhu:
## Description

Adding hindi as a lang pack


localize-behavior.js
```diff
@@ -69,6 +70,7 @@ D2L.PolymerBehaviors.Sequences.LocalizeBehaviorImpl = {
 					'fr': fr,
 					'fr-fr': frfr,
 					'fr-on': fron,
+					'hi':hi,
```
ianlaidlaw, line 73: super mega tiny nit: space after the `:`

ianlaidlaw (approved):

localize-behavior.js
GurpinderSandhu: i mega agree

brissons:
do we need to add something here as well (I am unfamiliar with serge)?
https://github.com/BrightspaceHypermediaComponents/sequences/blob/a91e77d89e6ebc3b2adbe6b85ebb25805436c7d0/d2l-sequences.serge.json

ianlaidlaw:
> do we need to add something here as well (I am unfamiliar with serge)? https://github.com/BrightspaceHypermediaComponents/sequences/blob/a91e77d89e6ebc3b2adbe6b85ebb25805436c7d0/d2l-sequences.serge.json

I _think_ that is just for those languages that have multiple versions... like handling fallbacks? I think if this is the base `hi` language and there's no other, there's nothing we need to do now. I think?

GurpinderSandhu:
> > do we need to add something here as well (I am unfamiliar with serge)? https://github.com/BrightspaceHypermediaComponents/sequences/blob/a91e77d89e6ebc3b2adbe6b85ebb25805436c7d0/d2l-sequences.serge.json
> 
> I _think_ that is just for those languages that have multiple versions... like handling fallbacks? I think if this is the base `hi` language and there's no other, there's nothing we need to do now. I think?

I would think so too, as serge was able to recognize this hi.js file and add a translation update PR to it already?

brissons (approved):

brissons:
Thanks!

PR #412, opened 2022-03-24
fixed sizing issue when nav bar is set to open

kristian-torres:
## Relevant Rally Links

[DE47637](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F628583327467&fdp=true) : LX > Spacing of Back to Content button on minibar

## Description

Problem - Current Behaviour

If the config variable d2l.Tools.Content.LearnerExperience.SideNavOpen is set to On (which results in the sidebar starting open), the spacing of the 'Back to Content' button and side navigation panel is wrong. If you close/reopen the side panel it is corrected.

## Goal/Solution

We fixed this in most scenarios in DE46051, looks like the one scenario of when the panel is already open needs to be fixed now.

## Video/Screenshots

https://user-images.githubusercontent.com/91631476/159953667-9f143d8e-ae30-4d40-8f31-20af579b94b4.mp4

## Remaining Work
- [ ] @Kiefer-Co-D2L will re-dev test. I was not able to consistently encounter this bug, but this should solve it.


d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -506,6 +506,7 @@ class D2LSequenceViewer extends mixinBehaviors([
 			this.mEntity = entity;
 			this._loaded = true;
 			if (entity.properties.sideNavOpen) {
+				this._sideBarSetInnerHeaderWidth(570);
```
ChrisKraljevicD2L, line 509: nit: could the `570` be made a `const`?

kristian-torres: whoops, I forgot to change that. Thanks

ChrisKraljevicD2L (approved):

coffeexcode (approved):

Kiefer-Co-D2L (commented):
Gave this one a dev test, still appears to happen on very large screen widths with that config var on:

image

ChrisKraljevicD2L:
> Gave this one a dev test, still appears to happen on large screen widths with that config var on:
> 
> image

interesting, the window width here isn't egregiously large given that the dev console is open. was it open before observing the defect or was it opened after?

kristian-torres:
Updated the check. I'll ask Kiefer to dev test this again.

PR #413, opened 2022-03-25
Remove unnecessary check causing ctrl click issues

connor-lowey:
## Relevant Rally Links

https://rally1.rallydev.com/#/289692574792d/teamboard?detail=%2Fdefect%2F628347821221&fdp=true

## Description

Clicking on the back button in sequences when holding the CTRL key would both open a new tab that would go back a page and send the current page back. The standard functionality is to open a new tab that would go back a page without changing the current tab.

## Goal/Solution

Remove unnecessary check for presumably sequences being in an iframe that will make the ctrl + click functionality more standard.

## Video/Screenshots
First example is chrome, second is firefox
DE47608


coffeexcode (approved):
Gave this a quick dev test. Tried the back to content button with a variety of clicks (left, middle, ctrl+left) and seems to do everything as expected!

brissons (approved):
lgtm!

ianlaidlaw (approved):

PR #418, opened 2022-04-01
DE47935: LX > End of Unit > Im Done/Back to Content button does not work

kristian-torres:
## Relevant Rally Links

[DE47935](https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F630486241523): LX > End of Unit > Im Done/Back to Content button does not work

## Description

In LX, at the end of unit page, clicking the Im Done/Back to Content button does not do anything.

## Goal/Solution

Readded a removed check to redirect the window to the sequence launcher but changed the check to include a check on the event that fired the _onClickBack() event.

## Video/Screenshots

https://user-images.githubusercontent.com/91631476/161321550-b5c68091-60df-4c6b-9ea4-88939beabf96.mp4

## Remaining Work


d2l-sequence-viewer/d2l-sequence-viewer.js
```diff
@@ -541,14 +541,19 @@ class D2LSequenceViewer extends mixinBehaviors([
...
-	_onClickBack() {
+	_onClickBack(event) {
 		this.telemetryClient.logTelemetryEvent('back-to-content');
 		this.dispatchEvent(new CustomEvent('back-to-content', { bubbles: true, composed: true }));
 
 		if (!this.backToContentLink) {
 			return;
 		}
 
+		if ((event && event.type !== 'click') && window.parent === window) {
```
ianlaidlaw, line 552: i think im a little confused about what this is doing, do you mind explaining?

we want to go back when the user clicks, but here we are checking if the event is not a click event? i might be missing something

kristian-torres: yeah, it's kinda weird. The tldr is we only want to enter this block when we are coming from the done/back to content button on the end of unit page. The only time the click event fires is from the top left header's back/back to content button which will skip over this to prevent both a new tab from opening and the current page from redirecting back to sequence launcher.

ChrisKraljevicD2L (approved):

PR #424, opened 2022-04-12
DE48099: Close new content dialog using keyboard

ianlaidlaw:
## Relevant Rally Links

- https://rally1.rallydev.com/#/289692574792d/custom/355050439968?detail=%2Fdefect%2F631328983127&fdp=true

## Description

The X button to close the new content alert would not work when attempting to close with the tab + enter buttons.

The reason this was happening was because of the manual `tabindex` being applied to the button, presumably to ensure they would be tabbed to after all other content on the page. This would change the natural appearance of the button being focused, but also removed keyboard functionality. These elements had a `tabindex` of `1` because we want the user to tab through content first before the new content alert.

Removing the `tabindex` kept the normal tab order but restored the missing style and keyboard functionality.

## Video/Screenshots

# BEFORE:

image

# AFTER:

image

newcontentfocus

## Remaining Work


coffeexcode (approved):
Pretty complex stuff, but solution LGTM

brissons (approved):

